PR #700, opened 2025-06-19
feat: OpenGUI load LIBs and SPEFs

### LINKED ISSUE #691: `OpenInOpenROAD` should load the SPEF if available
This allows to display the clock tree in the OpenROAD GUI.

So far I have implemented a hack in a fork of OpenLane which was very useful to debug the clock tree in [Greyhound](https://github.com/mole99/greyhound-ihp/tree/main)

Now that #688 has been merged, I'd like to implement this properly soon.

mole99:
`OpenROAD.OpenGUI`: The LIBs are now loaded by default and the SPEFs if available.
This makes it possible to view the clock tree, timing paths and the slack charts.

This is before SPEF extraction:

Bildschirmfoto vom 2025-06-19 17-12-40

And this after SPEF extraction:

Bildschirmfoto vom 2025-06-19 17-13-46

You can see how the clock tree is slightly less balanced and how some endpoints are violated because there is a huge load on a net (because of dont touch settings).

Also, I removed "Efabless Corporation" from authors in `pyproject.toml`.

Future improvements:

- Load the LIBs and SPEFs for all corners at once, instead of only the typical one. OpenROAD GUI allows to select the timing corner using a dropdown menu.

Edit: Should I merge this into main as it is a standalone feature, or into dev?

Fixes #691

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,9 @@
 name = "librelane"
 version = "2.4.0.dev7"
 description = "An infrastructure for implementing chip design flows"
+# Technically, maintainer. We cannot use the maintainers field until
+# poetry-core>=2.0.0 which requires Python version 3.9+. This field does
+# not imply primary authorship for the purposes of copyright.
 authors = ["Mohamed Gaber <me@donn.website>"]
 readme = "Readme.md"
 license = "Apache-2.0"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -449,9 +449,11 @@ jobs:
           path: current
       - name: "[PRs] Compare to target branch and comment result (if the workflow has secrets access)"
         uses: actions/github-script@v6
-        if: ${{ github.event_name == 'pull_request' }}
+        env: 
+            bot_github_token: ${{ secrets.BOT_GITHUB_TOKEN }}
+        if: github.event_name == 'pull_request' && env.bot_github_token != ''
         with:
-          github-token: ${{ secrets.BOT_GITHUB_TOKEN }}
+          github-token: ${{ env.bot_github_token }}
           script: |
             require("./.github/scripts/compare_metrics.js")({
               github: github,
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -1,8 +1,8 @@
 [tool.poetry]
 name = "librelane"
-version = "2.4.0.dev6"
+version = "2.4.1"
 description = "An infrastructure for implementing chip design flows"
-authors = ["Mohamed Gaber <me@donn.website>", "Efabless Corporation"]
+authors = ["Mohamed Gaber <me@donn.website>"]
```
donn, line 5: Not sure if we're legally allowed to do that. Efabless has to stay here for copyright purposes as they still own most of the codebase. This may change in the future, but not currently.

mole99: That's right, Efabless is the copyright holder for most of the code that is already there.
But I'm pretty sure this is not a copyright notice. If so, you'd have to add Sylvain, Uri, me... to this list.
The actual copyright holders are listed in the source files with the keyword "Copyright", that's the legal requirement. (You could also do a blanket copyright notice over the whole project or parts of it, but more granular is better.)

The authors in pyproject.toml are the active authors of this project I would say, and that's you.
Keeping Efabless there would send the wrong signals, as they are not active authors of LibreLane.

pyproject.toml
```diff
@@ -1,8 +1,8 @@
 [tool.poetry]
 name = "librelane"
-version = "2.4.0.dev6"
+version = "2.4.1"
```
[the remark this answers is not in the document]
mole99: Ah, thanks! The `.dev` notation is new for me ^^

mole99:
And one more thing: merging is currently blocked because the metrics cannot be uploaded. I can push my branch to upstream with no problem, but external contributors can't. Can we only run the last step if org==librelane?

pyproject.toml
donn: So `maintainers` is allowed (as well as many other pyproject.toml fields) with poetry 2.0.0 but that's Python 3.10+ exclusive. I think that's a fair sacrifice for LibreLane 3.0.0. For now I'll add a comment.

donn:
@mole99 It's not about signing off - all commits need to be cryptographically signed https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Though it will do that anyway when I, the repo admin, squashes. So no biggy.

PR #701, opened 2025-06-24
fix: dont load timing data of macros in OpenGUI

mole99:
I missed this one, sorry!

When a macro is part of the design, OpenInOpenROAD would attempt to load the spef of the macro, leading to a lot of warnings.
This doesn't work, even when the macro NL is loaded as OpenROAD seems to be flat by default and only sta is hierarchical. (Is that true?)

`gui.tcl` has been updated to only load the SPEF of the top-level design for the default corner.

Future improvements would be:
- Loading several corners at once
- Estimating routing delays if SPEF is not available

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -48,7 +48,7 @@ original authors after Efabless Corporation has ceased operations.
 
 * `OpenROAD.OpenGUI`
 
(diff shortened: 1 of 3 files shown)


donn:
id add a flag to read_spefs but its more important that main works for now

PR #710, opened 2025-07-02
Yosys Slang Integration

fkwilken:
Adds YOSYS_FRONTEND configuration variable that can select betwen Yosys, Slang, or Synlig frontends.


donn:
Thank you for the contribution -- this is already in `dev` as #702, though

PR #711, opened 2025-07-02
feat: flatten values of `List[Path]`

fkwilken:
Allows verilog files to be defined with multiple globs, for example:
```Yaml
VERILOG_FILES: 
  - dir::rtl/folder1/*.sv
  - dir::rtl/folder2/*.sv
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -64,6 +64,9 @@ original authors after Efabless Corporation has ceased operations.
 
 ## Misc. Enhancements/Bugfixes
 
+* `librelane.config.Variable`
+  * Variables of type `List[Path]` now flatten lists of lists so multiple globs
+    may be used within the same configuration variable.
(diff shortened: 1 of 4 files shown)


donn:
Hi, thanks for the contribution --

I agree with the need for this (it's been a pain on our side too), but I do not believe this is the way to implement it. This would cause a crash if the configuration lacks `VERILOG_FILES`, such as those for the [VHDLClassic](https://librelane.readthedocs.io/en/latest/reference/flows.html#vhdlclassic) flow (which uses `VHDL_FILES`.) And given that `VHDL_FILES` and all variables of `List[Path]` are affected, it may be prudent to find a more universal fix.

I've created https://github.com/librelane/librelane/issues/712 to track this issue - let me know if you'd like to tackle it. If not, I can look into it over the next week or so. Really sorry for the inconvenience!

PR #713, opened 2025-07-03
steps: In TclStep, overwrite env file, don't append

smunaut:
Append doesn't seem necessary and for steps that call multiple process, the env keeps growing indefinitely, up to 100M+ when you have lots of macros and slowing everything down.


mole99 (approved):
LGTM

PR #714, opened 2025-07-03
scripts/magic: Read LEF for any extra GDS that get included

smunaut:
Required to annotate the blocks with any position offset


mole99:
To add some background: the IHP IO cells had an offset in the LEF that is only considered when the LEF is read. The offset is now 0,0 again in the latest dev branch. Still, this fix is is needed for other cells that may have an offset.

I'm just wondering, if we should read the LEF every time we read a GDS, e.g. after the standard cells. This wouldn't be needed for all three open source PDKs, but is generally more correct. Does it slow down the flow measurable?

We can discuss this sometime, but this shouldn't block this PR.

PR #715, opened 2025-07-03
steps/klayout: More flexible option passing to DRC script

smunaut:
Uses the actual variable from the option instead of a fixed list. Just like in the LVS case.

cc @mole99

### DIFF
--- a/librelane/steps/klayout.py
+++ b/librelane/steps/klayout.py
@@ -445,12 +445,24 @@ class DRC(KLayoutStep):
         input_view = state_in[DesignFormat.GDS]
         assert isinstance(input_view, Path)
 
-        density_rules = (
-            "true"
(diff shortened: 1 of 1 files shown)


mole99:
Nice! We should actually do this and deduplicate code for all PDKs: no hardcoded settings, just configure the behavior in the PDK config. I just noticed there's no KLayout DRC setup for gf180mcuD, but there is a DRC ruleset in the PDK. We might do both at a later point.

Before merging this PR, for the max threads we should use `_get_process_limit()` as above in sky130, as this allows LibreLane to limit the number of threads per instance afaik.

smunaut:
Updated PR to use `_get_process_limit`

mole99:
I think something went wrong, I still see `os.cpu_count()`.

smunaut:
Oh yeah my bad ... I must have messed up between the rebase and the amend and the changes where staged but not commited ....

PR #719, opened 2025-07-09
feat!: KLayout.StreamOut: Add `KLAYOUT_CONFLICT_RESOLUTION`

mole99:
Add `KLAYOUT_CONFLICT_RESOLUTION` which specifies the conflict resolution if a cell name conflict arises.
Allowed values: "AddToCell", "OverwriteCell", "RenameCell" and "SkipNewCell"

Tested with an example design containing two different macros.
Using `SkipNewCell` there was only one copy of the stdcells. However, using `RenameCell` there were three copies: for the top-level and the two macros.

Edit: Update open_pdks hash to address issue in the OpenRAM macros.

@smunaut Feel free to give it a try

### DIFF
--- a/librelane/scripts/klayout/stream_out.py
+++ b/librelane/scripts/klayout/stream_out.py
@@ -162,7 +162,7 @@ def stream_out(
         print("[INFO] Checking for missing GDS…")
         missing_gds = False
         for i in top_only_layout.each_cell():
-            if i.is_empty():
+            if i.is_ghost_cell():
                 missing_gds = True
                 print(
                     f"[ERROR] LEF Cell '{i.name}' has no matching GDS cell.",


Changelog.md
```diff
@@ -18,6 +18,12 @@
 
 ## Steps
 
+* `KLayout.StreamOut`
+
+  * Added `KLAYOUT_CONFLICT_RESOLUTION` which specifies the conflict resolution if a cell name conflict arises.
+  
+    * Allowed values: "AddToCell", "OverwriteCell", "RenameCell" and "SkipNewCell"
```
smunaut, line 25: Maybe add a note about the default value ?

librelane/steps/openroad.py
```diff
@@ -2730,7 +2730,7 @@ class OpenGUI(OpenSTAStep):
 
     inputs = [
         DesignFormat.ODB,
-        DesignFormat.SPEF.mkOptional(),
+        # DesignFormat.SPEF.mkOptional(),
```
smunaut, line 2733: Not sure this is meant to be in this commit ?

mole99:
Alright, so it seems when the top-level DEF is loaded, all references to other cells (except for VIAs created by OpenROAD) such as standard cells and macros should become so-called "ghost cells", as they are called in KLayout.
They are identified by a parantheses around their cell name, e.g. "(sky130_ef_sc_hd__fill_1)".

By loading the external GDS files these ghost cells are resolved and link to the actual cell.

Now it seems for "sky130_sram_1kbyte_1rw1r_32x256_8" no ghost cell is created, instead it's just an empty cell.

With the previous default behavior of "AddToCell" KLayout would just add the content of the actual "sky130_sram_1kbyte_1rw1r_32x256_8" cell when the GDS is loaded. But with the new default behavior of "RenameCell", the actual cell with the content is renamed to "sky130_sram_1kbyte_1rw1r_32x256_8$1", therefore the sram cell in the top-level cell stays empty.

The new behavior would be the more correct one, but first I need to find out why "sky130_sram_1kbyte_1rw1r_32x256_8" does not create a ghost cell.

smunaut:
Well the default behavior of `magic` doesn't match the default behavior of `KLayout` so that yields XOR differences.

In TT in particular we definitely need the `RenameCell` behavior ( which is equivalent to what `magic` does ) because different projects could use different version of the cell library and when combining them it's important that each project keeps using their respective version of the cell since they might not be compatible ...

Ideally there would be some behavior that checks if the different loaded version are the same and if they are, then only keep one and only rename if loading a different version, but that's not a behavior that exists.

mole99:
Thanks, it took a while 😄

open_pdks [PR](https://github.com/RTimothyEdwards/open_pdks/pull/483) is open, now just need to wait for the merge before creating a new ciel release.

mole99:
So the open_pdks update fixed the issue with the SRAM as expected, however it uncovers a separate (pre-existing) issue:

Two of the test cases fail, `aes_upw_old` and `io_place`. The problem lies in the XOR step: the `conb` cell shows differences between magic and KLayout. This cell was updated in the PDK some time ago, `npc.drawing` was extended to prevent issues when abutting other cells.

Both of the tests use macros that still include the old version of the cell. With this PR, KLayout correctly preserves the old cell in the macros, however magic replaces the conb cell in the macros with the new version of the PDK.
This differs from the expected behavior: magic should create duplicates for all cells of the same name.

If I set `KLAYOUT_CONFLICT_RESOLUTION` back to the previous implicit default value of `AddToCell`, then there are no XOR differences. KLayout will merge the conb cell in the macros with the one in the PDK, and since the change is only an extension of a layer, the resulting cell is effectively the new cell from the PDK. (But it only hides the issue.)

---

I would propose to merge this PR as it is, as it adds a new feature that we need and uncovers an issue that already exists. I will open a proper "Issue" for it then. (We would need to disable the two test cases for now :/ )

If a user encounters the error, there are two ways to proceed:

- Reharden the macros with the new PDK
- Use the GDS from KLayout

But at least there is a notification about an issue.

mole99:
Well, I can send you my local run for `io_placer`: [RUN_2025-07-22_09-55-28.zip](https://github.com/user-attachments/files/21364186/RUN_2025-07-22_09-55-28.zip)
It was done with this PR and automatic PDK.

Basically, it places two macros in the `user_project_wrapper` and the pins are placed using `pin_order.cfg`:

image

smunaut:
Ah, it's because in that test `MACROS` is `None` and it uses the "old" style of using EXTRA_GDS and EXTRA_LEFS and such ... and so it doesn't follow the same processing path.

mole99:
All green! Going to merge as discussed.

PR #722, opened 2025-07-10
feat: Add support for nested dicts in tcl

mole99:
Allows to create nested dicts, as required in the PDK config for:

- the `LAYERS_RC` dict
- the `VIAS_R` dict
- the `LIB` dict

In preparation for upstreaming support for the IHP PDK.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -222,6 +222,10 @@
 
 ## Misc. Enhancements/Bugfixes
 
+* `librelane.common`
+
+  * `_eval_env`: Add support for nested dicts in tcl
+
 * `openlane.flows`
 
   * `SequentialFlow`
--- a/librelane/common/tcl.py
+++ b/librelane/common/tcl.py
@@ -1,3 +1,7 @@
+# Copyright 2025 LibreLane Contributors
+#
+# Adapted from OpenLane
+#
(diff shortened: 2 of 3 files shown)


donn:
Hiya Leo- slight concern about this, as I recall `dict set` does not work like the Tcl `dict set` which may confuse people trying to get their own PDKs to be compatible.

It's ultimately not a huge issue but it's another thing we'd have to add a warning for in the documentation, when we can just make the new command `set_dict_key_path` or something and sidestep the entire issue. But you have to propagate these changes to the IHP Open PDK proper and it quickly spirals…

What do you think?

mole99:
Thanks for taking a look!

Actually, `dict set` should behave the same as the Tcl `dict set`. I had to update it for nested dicts in order to support `LAYERS_RC` and `VIAS_R`.
However, none of the other dozen sub-commands are supported, only `set`.

I think the way the Tcl Python interface is implemented by intercepting commands makes it hard to reach 100% compatibility. Another approach could be to run the interpreter and convert all env variables to Python at the end. Not sure if that's possible 😅

mole99:
Maybe we can revisit this at some point and see what the issues are.

But for now, I think it's okay to only support `dict set`?

PR #725, opened 2025-07-17
[WONTFIX] Fixed tilde (~) not expanding

avgsurfman:
### Bug description
Librelane fails to expand the home path with a tilde (~) before passing the config down to the config function.

Upon a short inspection with pdb, it turned out that right before the failure, the path didn't expand right before
it hits the InvalidConfig exception.
This has been narrowed down due to
- `PDK_ROOT=` environment variable 
- "$HOME" 
working as intended.

This is a one-liner fix.
```[nix-shell:~/librelane]$ librelane --manual-pdk --pdk-root=~/.ciel --pdk ihp-sg13g2 ../my_designs/pm32-example/config.json
> /home/de_generic/librelane/librelane/__main__.py(90)run()
-> pdk_root = os.path.expanduser(pdk_root) if isinstance(pdk_root, str) else None
(Pdb) locals()
{'ctx': <cloup._context.Context object at 0x7bffd6b0af30>, 'flow_name': None, BEFORE->>>> 'pdk_root': '~/.ciel', 'pdk': 'ihp-sg13g2', 'scl': None, 'config_files': ('../my_designs/pm32-example/config.json',), 'tag': None, 'last_run': False, 'frm': None, 'to': None, 'skip': (), 'overwrite': False, 'reproducible': None, 'with_initial_state': None, 'config_override_strings': (), '_force_run_dir': None, 'design_dir': None, 'initial_state_element_override': (), 'view_save_path': None, 'ef_view_save_path': None, 'TargetFlow': <class 'librelane.flows.classic.Classic'>, 'config_file': '../my_designs/pm32-example/config.json'}
(Pdb) continue
> /home/de_generic/librelane/librelane/__main__.py(92)run()
-> if meta := Config.get_ueta(config_file):
(Pdb) locals()
{'ctx': <cloup._context.Context object at 0x7bffd6b0af30>, 'flow_name': None, AFTER->>>> 'pdk_root': '/home/de_generic/.ciel', 'pdk': 'ihp-sg13g2', 'scl': None, 'config_files': ('../my_designs/pm32-example/config.json',), 'tag': None, 'last_run': False, 'frm': None, 'to': None, 'skip': (), 'overwrite': False, 'reproducible': None, 'with_initial_state': None, 'config_override_strings': (), '_force_run_dir': None, 'design_dir': None, 'initial_state_element_override': (), 'view_save_path': None, 'ef_view_save_path': None, 'TargetFlow': <class 'librelane.flows.classic.Classic'>, 'config_file': '../my_designs/pm32-example/config.json'}```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,10 @@
 ## Documentation
 -->
 
+# 2.4.2
+
+* Fixed tilde (~) not being expanded for home directory
+
 # 2.4.1
 
 ## Misc. Enhancements/Bugfixes
--- a/librelane/__main__.py
+++ b/librelane/__main__.py
@@ -83,6 +83,10 @@ def run(
         TargetFlow: Optional[Type[Flow]] = Flow.factory.get("Classic")
 
         for config_file in config_files:
+            # expands the pdk-root path
+            pdk_root = (
+                os.path.expanduser(pdk_root) if isinstance(pdk_root, str) else None
+            )
             if meta := Config.get_meta(config_file):
                 # to maintain backwards compat, in 3 you will need to explicitly
                 # set the flow you're substituting
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "librelane"
-version = "2.4.1"
+version = "2.4.2"
 description = "An infrastructure for implementing chip design flows"
 # Technically, maintainer. We cannot use the maintainers field until
 # poetry-core>=2.0.0 which requires Python version 3.9+. This field does

### CODE REVIEW

**Comment by mole99** [comment]
Hi, sorry this took so long!

The change looks good to me, but we need to rebase it. @donn I think it would make sense to rebase it on top of main instead of dev since it is a bug fix. What do you think?

@avgsurfman If you want I can take care of that for you.

**Comment by avgsurfman** [comment]
The only change that I forgot to push is that I expanded the single line into a cleared if else statement instead.
I've been reading Slatkin's Effective Python, and it's generally better to avoid conditional expressions if they're too long.
This really doesn't matter but FYI this is why I pushed last minute. I think It will pass the checks.

**Comment by mole99** [comment]
@Donn Can you tell me why this PR was deployed to PyPI 😶

**Comment by donn** [comment]
Apologies for taking so long to respond to this!

---

`~` is in general a shell convenience and it's the shell's responsibility to expand it, not underlying tools, and its appearance in variables is generally a bad thing because even common UNIX shell tools don't know how to deal with it:

```sh
~% X=~/.pdk; Y='~/.pdk'; 
~% echo $X $Y
/Users/donn/.pdk ~/.pdk
~% mkdir $Y
mkdir: ~: No such file or directory
```

You'll see here that for X, the shell expanded the `~` before setting the value, but for Y, as a literal string, it didn't. That's why it's more typical for people to use `$HOME` in their `.profile`/`rc` files.

So I'm hesitant to merge this because it would enable a practice that really shouldn't be done in the first place. Our CLI library, Click, which [we use to handle the variable `--pdk-root` (including the environment as a fallback)](https://github.com/librelane/librelane/blob/a8ca5846680109e8f8a9215a477e3001220af224/librelane/flows/cli.py#L353) also disapproves of this:

> resolve_path ([bool](https://docs.python.org/3/library/functions.html#bool)) – Make the value absolute and resolve any symlinks. A `~` is not expanded, as this is supposed to be done by the shell only.
>
> from https://click.palletsprojects.com/en/stable/api/#click.Path

Could you elaborate on if there is a particular setup that necessitates this?

avgsurfman:
I'm using zsh if that helps.

avgsurfman:
I don't think it's the "intended" behavior as unless someone's bash-fu is expert-level, tilde (~) and $HOME are pretty much equivalent unless you're writing a bash script (not user-level). I'm not writing $HOME every time I'm writing cd in my terminal.
At the very least, the error handling messages should be clearer. If lLL returned a string with the full path where the program failed in the error message, it would be *easier* for the user to figure out what's wrong.
The error message doesn't print the resolved path in a way it would make it comprehensive for the user to figure out that they used a wrong symbol, unless the prerequisite to use librelane is to read the bash HTML manual from cover to cover.

PR #739, opened 2025-08-05
fix: incompatibility between #688 and #612

donn:
Fixes a minor incompatibility between #688 and #612 that causes a crash when TclSteps have a missing optional input.

In `main`, `[]` is guaranteed to return a result - `None`. So the error didn't bubble up.


donn:
Merging even with the container issue since that is in `dev` as well. Will need to fix that tomorrow.

Issue #745, opened 2025-08-13
Yosys Slang argument propagation


towoe:
### Description

I was trying to use the Slang frontend in the Yosys script and keep the hierarchy.
The option `SYNTH_HIERARCHY_MODE` is set to `keep` but this has no effect.
Looking at the code I think this is due to the Yosys Slang frontend not getting anything "special" set in 

https://github.com/librelane/librelane/blob/948d921862e2ac0e1e8b56322bceaf6cbfe56238/librelane/scripts/pyosys/ys_common.py#L66-L76

yosys-slang provides the option `--keep-hierarchy`
https://github.com/povik/yosys-slang/blob/fd77b320b865ee75560edab4258ced1e1dd47ecf/src/slang_frontend.cc#L44-L45
When this option is used directly in `ys_common.py` the hierarchy is preserved.

Further link to Slang options: https://sv-lang.com/command-line-ref.html

### Proposal

I think there are two ways to do it.
Either access `SYNTH_HIERARCHY_MODE` in `ys_common.py` and add `--keep-hierarchy` to the yosys-slang call.
Or create a new variable for a list of options to yosys-slang for forwarding of options.

Sorry if I missed something here, still a bit new to librelane.
Thanks for this awesome tool :)

towoe:
And now I see PR #726 ...
So I guess this will fix this issue

mole99:
I'm not sure if the PR #726 fixes your request, it does not pass `--keep-hierarchy` to yosys-slang from what I see.

But I created #747 that you can use in the meantime.

towoe:
Thanks @mole99!
Yeah, I was mostly referring to #726 as I assumed this will change a few things, and maybe this is then also a good place to think about the argument passing.
But thanks for jumping on it and providing the PR, much appreciated!

PR #747, opened 2025-08-14
feat: Add `SLANG_ARGUMENTS` to pass arguments to Slang

mole99:
This adds argument passing to Slang as requested by #745.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,8 @@
 
 ## Steps
 
+* `Yosys.Synthesis`: Added `SLANG_ARGUMENTS`, which is used to pass arguments to the Slang frontend.
+
   * `KLayout.StreamOut`: Added `KLAYOUT_CONFLICT_RESOLUTION` which specifies the conflict resolution if a cell name conflict arises. (Default: "RenameCell")
 
     * Allowed values: "AddToCell", "OverwriteCell", "RenameCell" and "SkipNewCell"
--- a/librelane/scripts/pyosys/json_header.py
+++ b/librelane/scripts/pyosys/json_header.py
@@ -63,6 +63,7 @@ def json_header(
         includes=includes,
         defines=defines,
         use_slang=config["USE_SLANG"],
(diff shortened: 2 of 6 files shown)


towoe:
Thanks again for working on this @mole99!
This definitely allows full control for the user.
The only concern I would have is that it might get confusing with the different options available.
For example, looking at `SYNTH_HIERARCHY_MODE`, I would not instantly know that I also have to set something manually in `SLANG_ARGUMENTS` to make this work for the Slang plugin as well.
The solution might be more notes in the options which have a link to the Yosys Slang plugin.
Or it could also be that there is an additional check in the `USE_SLANG` case, updating the arguments from related options.

mole99:
You're welcome!

I don't entirely agree with the last point, in my view those are two different mechanisms that should be controlled separately (or at least you should have the option). For some strange reason, you might want to run Slang hierarchically and then flatten later in Yosys (there is actually a use case for that, see inout ports).

But I agree that it might be confusing, especially since it's called `SYNTH_HIERARCHY_MODE` and not `YOSYS_HIERARCHY_MODE`.

I'm open to suggestions, but adding a note to `SYNTH_HIERARCHY_MODE` regarding Slang is definitely a good idea :)

mole99:
I see the value in the abstraction, but doing it this way would already be a problem for my design where I need to:

1. keep the hierarchy in Slang
2. use `keep_hierarchy` in Yosys on instances above a certain gate count
3. flatten the design with Yosys

For the second point I was thinking about adding a variable to LibreLane (e.g. `SYNTH_HIERARCHY_GATE_THRESHOLD`) that sets the gate count threshold, but then we would need another `SYNTH_HIERARCHY_MODE` (e.g. `partial`) that does everything above (but it would only be needed because of Slang). With `SLANG_ARGUMENTS` we could just add `SYNTH_HIERARCHY_GATE_THRESHOLD`. Also, we need `SLANG_ARGUMENTS` anyways as you can specify many other things.

PR #753, opened 2025-08-27
feat: Add support for user defined corner analysis during STA

crispy245:
This PR introduces a new config named `STA_EXTRA_CORNER_TCL_FILE` that
allows users to define extra tests to be ran during STA executions.

In short it, during the STA operations librelane calls `corner.tcl`. Now
`corner.tcl` calls a user defined `.tcl` file to be ran alongside any
pre-existing checks.

---

Example:
```
{
  "DESIGN_NAME": "barret_red",
  "VERILOG_FILES": ["dir::src/*v"],
  "CLOCK_PERIOD": 8,
  "CLOCK_PORT": "i_clk",
  "STA_EXTRA_CORNER_TCL_FILE":"dir::my_own_config.tcl"
}
```

> my_own_config.tcl
```
puts "%OL_CREATE_REPORT my_own_max.rpt"
puts "\n==========================================================================="
puts "report_checks -path_delay max (Setup)"
puts "============================================================================"
puts "======================= [$corner name] Corner ===================================\n"
report_checks -sort_by_slack -path_delay max -fields {slew cap input nets fanout} -format full_clock_expanded -group_count 1000 -corner [$corner name]
puts ""
puts "%OL_END_REPORT"
```

ccing @mole99 since he helped me with some ideas


donn:
Can you re-base your changes against the `dev` branch? We've released, so `main` PRs are only for new versions and critical bugfixes.

librelane/steps/openroad.py
```diff
@@ -258,6 +258,11 @@ class OpenROADStep(TclStep):
             Optional[Path],
             "Points to the DEF file to be used as a template.",
         ),
+        Variable(
+            "STA_EXTRA_CORNER_TCL_FILE",
+            Optional[Path],
+            "Specifies a additional configuration .tcl file to be called during (PnR) steps.",
```
donn: In the interest of time, we will merge this marking it as experimental and unstable. The thing is, and the reason we haven't merged this so far is that we can't commit to the same variable names always working for example. This is already a bit of a problem with SDC files and PDNs, and it will be a problem here. But until we figure out a solution, this should be fine.

donn (approved):

PR #758, opened 2025-09-02
feat: `Magic.SpiceExtraction`: replace `MAGIC_NO_EXT_UNIQUE` with enum `MAGIC_EXT_UNIQUE`

mole99:
Allowed values are: "all", "notopports", "noports", "none"

This allows for situations where only the top ports should be connected by label, or the other way around.

---

Depends on https://github.com/librelane/librelane-step-unit-tests/pull/2

### DIFF
--- a/librelane/steps/magic.py
+++ b/librelane/steps/magic.py
@@ -458,15 +458,18 @@ class SpiceExtraction(MagicStep):
         Variable(
             "MAGIC_EXT_ABSTRACT_CELLS",
             Optional[List[str]],
-            "A list of regular experssions which are matched against the cells of a "
+            "A list of regular expressions which are matched against the cells of a "
             + "the design. Matches are abstracted (black-boxed) during SPICE extraction.",
         ),
         Variable(
-            "MAGIC_NO_EXT_UNIQUE",
-            bool,
-            "Enables connections by label in LVS by skipping `extract unique` in Magic extractions.",
-            default=False,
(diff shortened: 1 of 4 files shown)


librelane/steps/magic.py
```diff
@@ -458,15 +458,15 @@ class SpiceExtraction(MagicStep):
...
-            "MAGIC_NO_EXT_UNIQUE",
-            bool,
-            "Enables connections by label in LVS by skipping `extract unique` in Magic extractions.",
-            default=False,
-            deprecated_names=["LVS_CONNECT_BY_LABEL"],
+            "MAGIC_EXT_UNIQUE",
+            Literal["all", "notopports", "noports", "none"],
+            'Runs `extract unique` with the specified option. The default is "all", and "none" disables `extract unique`, allowing connections between separate nets by label in LVS.',
+            default="all",
+            deprecated_names=["MAGIC_NO_EXT_UNIQUE", "LVS_CONNECT_BY_LABEL"],
```
donn, line 469: So, for `deprecated_names` where the type and logical value have been changed, you need to provide the mapping explicitly like this:

```
deprecated_names=[
  ("MAGIC_NO_EXT_UNIQUE", lambda o: "none" if o else "all"),
  ("LVS_CONNECT_BY_LABEL", lambda o: "none" if o else "all"),
]
```

mole99:
CI is green! I will merge the steps and then update the submodule.

PR #760, opened 2025-09-02
feat: Yosys.Synthesis: Added variables to set `keep_hierarchy`

mole99:
* `SYNTH_KEEP_HIERARCHY_MIN_COST`: Sets the `keep_hierarchy` attribute on modules where the gate count is estimated to exceed the specified threshold. This prevents larger modules from being flattened.
* `SYNTH_KEEP_HIERARCHY_INSTANCES`: A list of instances for which to set the `keep_hierarchy` attribute.
* `SYNTH_KEEP_HIERARCHY_MODULES`: A list of modules for which to set the `keep_hierarchy` attribute.

Two things are important to note:

- In `check_macro_instances.tcl` I had to remove `-hierarchical` on `get_cells` since this would search for the specified cell in each hierarchy level. This works fine when everything is flattened, but doesn't when you have a hierarchy and specify the hierarchical path to a cell, i.e., `a/b/c`.
- I had to set the `keep_hierarchy` attribute on all blackbox modules, since `keep_hierarchy -min_cost` would otherwise simply stop when encountering a bb module. (The other option would be to set a min_cost.)


mole99:
`SYNTH_KEEP_HIERARCHY_MIN_COST` and `SYNTH_KEEP_HIERARCHY_MODULES` worked both fine when I tested them in Greyhound. I couldn't get `SYNTH_KEEP_HIERARCHY_INSTANCES` to work, but I suppose this is due to me using yosys-slang which changes a lot of the naming. The only difference to `SYNTH_KEEP_HIERARCHY_MODULES` is the omission of `-mod`.

mole99:
@Donn If you agree with the variable names, this would be ready to merge.

mole99:
Rebased and updated changelog.

PR #765, opened 2025-09-03
feat: `Verilator.Lint`: Allow disabling of warnings

mole99:
* `Verilator.Lint`

  * Added `LINTER_DISABLE_WARNINGS` to disable linter warnings.

  * Added `LINTER_DISABLE_WARNINGS_BLACKBOX` to disable linter warnings for blackbox modules.

  * Added `LINTER_VLT` as a user defined Verilator Configuration format file (`.vlt`).

There is definitely potential for discussion.

For example, I have set the default values of `LINTER_DISABLE_WARNINGS` and `LINTER_DISABLE_WARNINGS_BLACKBOX` to the previous defaults. This means you need to specify those as well when you override these variables, but it also allows you to remove them if you wish.

I also added the option to supply a user defined Verilator Configuration format file. It must to be read in before the source files to take effect, so I moved `extra_args` up.


mole99:
Now I understand, those are specific warnings for each source file!

Thanks for the idea, added and works great:

```
// DESCRIPTION: Verilator output: Waivers generated with --waiver-output

`verilator_config

// Below are suggested waivers. You have three options:
//   1. Fix the reason for the linter warning in the Verilog sources
//   2. Keep the waiver permanently if you are sure it is okay
//   3. Keep the waiver temporarily to suppress the output

// lint_off -rule UNUSEDSIGNAL -file "*/home/leo/Repositories/librelane-prs/test/designs/spm/src/spm.v" -match "Signal is not driven, nor used: 'xy'*"

```

PR #772, opened 2025-09-20
feat: `Magic.DRC`: Added `MAGIC_GDS_FLATGLOB`

mole99:
* Used to flatten cells in order to prevent false positive DRC errors.


mole99:
This one is rather harmless. I would merge it tomorrow if there are no objections :)

PR #786, opened 2025-10-02
feat: add `MAGIC_DRC_MAGLEFS` to blackbox cells during DRC

mole99:
This PR adds `MAGIC_DRC_MAGLEFS` to load a list of maglefs before reading the actual layout, in order to blackbox certain cells.

Note: There are already two existing variables as part of `MagicStep`: `CELL_MAGS` and `CELL_MAGLEFS`
However, those are only used as fallback for (blackboxed) cells. In addition, it seems their functionality **isn't implemented** anymore?

If so, we should either remove them, implement their functionality or merge them in some way with `MAGIC_DRC_MAGLEFS`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -62,6 +62,10 @@ Style Notes
 
   * Made `DesignFormat.DEF` optional
 
+  * Added `MAGIC_DRC_MAGLEFS`
+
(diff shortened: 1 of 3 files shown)


mole99:
Me neither, that's why I have Tim 😁

PR #787, opened 2025-10-03
feat: While Step

KelvinChung2000:
Adding a new step construct, which can be used for optimisation. For example, finding the smallest die area.

### DIFF
--- a/librelane/steps/step.py
+++ b/librelane/steps/step.py
@@ -13,72 +13,73 @@
 # limitations under the License.
 from __future__ import annotations
 
-import datetime
-import json
 import os
-import shutil
-import subprocess
 import sys
-import textwrap
+import json
 import time
-from abc import ABC, abstractmethod
-from concurrent.futures import Future
-from decimal import Decimal
-from inspect import isabstract
-from io import TextIOWrapper
-from itertools import zip_longest
+import psutil
+import shutil
+import textwrap
+import datetime
+import subprocess
 from signal import Signals
+from decimal import Decimal
+from io import TextIOWrapper
 from threading import Thread
+from inspect import isabstract
+from itertools import zip_longest
+from abc import abstractmethod, ABC
+from concurrent.futures import Future
 from typing import (
     Any,
-    Callable,
-    ClassVar,
-    Dict,
-    Generic,
     List,
+    Callable,
     Optional,
-    Sequence,
     Set,
-    Tuple,
-    Type,
-    TypeVar,
     Union,
+    Tuple,
+    Sequence,
+    Dict,
+    ClassVar,
+    Type,
+    Generic,
+    TypeVar,
 )
 
-import psutil
 from rich.markup import escape
 
-from .. import logging
-from ..__version__ import __version__
-from ..common import (
-    GenericDict,
-    GenericDictEncoder,
-    GenericImmutableDict,
-    Path,
-    RingBuffer,
-    Toolbox,
-    copy_recursive,
-    final,
-    format_elapsed_time,
-    format_size,
-    mkdirp,
-    protected,
-    slugify,
-)
 from ..config import (
     Config,
     Variable,
     universal_flow_config_variables,
 )
+from ..state import DesignFormat, State, InvalidState, StateElement
+from ..common import (
+    GenericDict,
+    GenericImmutableDict,
+    GenericDictEncoder,
+    Toolbox,
+    Path,
+    RingBuffer,
+    mkdirp,
+    slugify,
+    final,
+    protected,
+    copy_recursive,
+    format_size,
+    format_elapsed_time,
+)
+from .. import logging
 from ..logging import (
-    debug,
-    err,
-    info,
     rule,
     verbose,
+    info,
     warn,
+    err,
+    debug,
 )
-from ..state import DesignFormat, InvalidState, State, StateElement
+from ..__version__ import __version__
+
 
 VT = TypeVar("VT")
 
@@ -721,9 +722,8 @@ class Step(ABC):
 
             IPython.display.display(IPython.display.Markdown(Self.get_help_md()))
         except NameError:
-            from rich.markdown import Markdown
-
             from ..logging import console
+            from rich.markdown import Markdown
 
             console.log(Markdown(Self.get_help_md()))
 
--- a/librelane/steps/step.py
+++ b/librelane/steps/step.py
@@ -48,8 +48,6 @@ from typing import (
 import psutil
 from rich.markup import escape
 
-from librelane.flows.flow import FlowProgressBar
-
 from .. import logging
 from ..__version__ import __version__
 from ..common import (
@@ -72,6 +70,7 @@ from ..config import (
     Variable,
     universal_flow_config_variables,
 )
+from ..flows.flow import FlowProgressBar
 from ..logging import (
     debug,
     err,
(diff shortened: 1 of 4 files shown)


librelane/steps/step.py
```diff
@@ -13,73 +13,73 @@
 # limitations under the License.
 from __future__ import annotations
 
-import os
```
donn, line 16: Our import orders are currently "length then alphabetically" for all source files. It's not ideal but changing it for one file is not very consistent either.

librelane/steps/step.py
```diff
@@ -635,7 +635,7 @@ def get_help_md(
                 %s
                 ```
 
-                {':::{dropdown} Importing' if use_dropdown else '#### Importing'}
+                {":::{dropdown} Importing" if use_dropdown else "#### Importing"}
```
donn, line 638: If this doesn't break Python versions below 3.12 I'm okay with it

donn (changes requested):
Overall, great idea! Some changes, but overall pretty positive.

I do have a minor concern re: FlowProgressBar: it's really not designed to be used outside flows and would take its own PR to rework it so it can work with steps too (and have things like multiple stacked progress bars… etc.) I think you can leave this out of this PR.

Also, if it's not too much of a bother, a test for this in https://github.com/librelane/librelane/dev/main/test/steps/test_step.py would be appreciated — you can use a dummy step that just increments some kind of metric to just test that While steps work as expected.

donn (approved):
LGTM- @mole99 an architectural double-check would be nice on this

(+ @kareefardi if you have a moment lmao)

mole99 (changes requested):
Great work! I have a few nitpicks:

- Why is the ruff section needed? LibreLane uses black for parsing and formatting.
- Ruff seems to have reordered a number of import statements and replaced some string quotation marks, which adds to the noise.
- How about allowing `None` for `max_iterations` to run until the condition is met?
- It would be great to document the step under: https://librelane.readthedocs.io/en/latest/usage/writing_custom_flows.html

mole99:
The implementation looks good, and it looks like the While step can be very useful for designing flows more expressively.

As I understand it, right now you have to define the break condition as a function, and the last state is always used?
This can be used in a scenario where you want to find the smallest `DIE_AREA`. You continuously shrink `DIE_AREA` until the flow throws an exception. The state of the last executed step is the one you want to use.

However, I think it could also be used in different ways: You supply a number of values for a variable: e.g. `SYNTH_STRATEGY = ["AREA 0", "AREA 1", ...]`, and  specify a metric that you want to optimize (minimum, maximum) e.g. the area. The while step would then run the sub-flow for all variable values and select the state with the best metric.

This would simplify the `SynthesisExploration` flow. But you can easily create new flows where you want to use different variable values.
It may even be possible to supply several variables and their lists of values and a mode how to combine those values (e.g. one after another or every value with every value).

Maybe it can be integrated in the While step? Perhaps this would be better suited as a separate "For" step? But the While and For step could probably share some code.

The idea is that, if we only need to supply the sub-flow, the values for the variable, and the metric to optimize, we could even create such flows through the config file, or through the top-level API of e.g. SequentialFlow.

These are all suggestions. What do you think?

donn:
@mole99 I added ruff. Ruff has an advantage in that it can detect elements that are unparseable in Python 3.10 even if you're running it under Python 3.12. So that's on me.

The rest should be addressed, and I only have Ruff detecting parsing errors though, so I don't know what happened there.

KelvinChung2000:
> * How about allowing `None` for `max_iterations` to run until the condition is met?

I think `None` is a bad idea, as most of the time we don't want it to run forever. 10 is a reasonable starting point. The user should override it anyway to meet their needs. For example, add a config variable that, at the start of `run`, reads and modifies `max_iterations`.   

> * It would be great to document the step under: https://librelane.readthedocs.io/en/latest/usage/writing_custom_flows.html

I will add some documentation later.

> As I understand it, you have to define the break condition as a function, and the last state is always used?
> This can be used in a scenario where you want to find the smallest DIE_AREA. You continuously shrink DIE_AREA until the flow throws an exception. The state of the last executed step is the one you want to use.

The last state before the condition breaks is used (the end state of the previous iteration), but you can also modify it with `post_loop_callback` to apply extra changes. For the cases you suggest, the user should store the result using post-iteration callback after each iteration into a step variable, then use post-loop callback to recover the state. 


> However, I think it could also be used in different ways: You supply a number of values for a variable: e.g. SYNTH_STRATEGY = ["AREA 0", "AREA 1", ...], and specify a metric that you want to optimize (minimum, maximum) e.g. the area. The while step would then run the sub-flow for all variable values and select the state with the best metric.
> 
> This would simplify the SynthesisExploration flow. But you can easily create new flows where you want to use different variable values.
> It may even be possible to supply several variables and their lists of values and a mode how to combine those values (e.g. one after another or every value with every value).


This is supported by providing a mix of pre-iteration callback and post-iteration callback.

> This would simplify the SynthesisExploration flow. But you can easily create new flows where you want to use different variable values.
> It may even be possible to supply several variables and their lists of values and a mode how to combine those values (e.g. one after another or every value with every value).
> 
> Maybe it can be integrated in the While step? Perhaps this would be better suited as a separate "For" step? But the While and For step could probably share some code.

I would say the "for" loop will essentially be an extension of the while loop with a bit more constraints. 

> The idea is that, if we only need to supply the sub-flow, the values for the variable, and the metric to optimize, we could even create such flows through the config file, or through the top-level API of e.g. SequentialFlow.

I don't think this is suitable as a top-level flow API, as it is too complex and too much is configurable. If you're doing optimisation, you want a lot more control over it and to be touching the Python yourself. Defining a system to parse conditions also feels painful to implement.

mole99:
> I think None is a bad idea, as most of the time we don't want it to run forever. 10 is a reasonable starting point. The user should override it anyway to meet their needs. For example, add a config variable that, at the start of run, reads and modifies max_iterations.

Well, then the While step is actually a For step with a break condition ^^

I'm not too fussed about this, you can always set `max_iterations` to 100000. Still, I think it would be more generic this way.

> This is supported by providing a mix of pre-iteration callback and post-iteration callback.

Yes, all of that can be implemented using callbacks. My suggestion was to provide a concrete implementation for it so people can use it straight away.

This doesn't have to be in this PR. I wanted to hear what you are thinking.

> I don't think this is suitable as a top-level flow API, as it is too complex and too much is configurable. If you're doing optimisation, you want a lot more control over it and to be touching the Python yourself. Defining a system to parse conditions also feels painful to implement.

I think the use case of iterating over a number of values of a variable could very well be configured at the top-level.

The Classic flow is defined as: https://github.com/librelane/librelane/blob/9c6deea5923e65eff390942f3ba941da5053681c/librelane/flows/classic.py#L31C7-L31C14

Let's say, we allow something like this:

```
    Steps = [
        Verilator.Lint,
        Checker.LintTimingConstructs,
        Checker.LintErrors,
        Checker.LintWarnings,
        Yosys.JsonHeader,
        Common.For,
        [
            Yosys.Synthesis,
            Checker.YosysUnmappedCells,
            Checker.YosysSynthChecks,
            Checker.NetlistAssignStatements,
            ...
        ]
        ...
```

The metric to optimize and the variables to iterate over can be specified as variables of the For step:

```
FOR_METRIC = "design__die__area"
FOR_OPTIMIZATION = "minimum"
FOR_VARIABLES:
  SYNTH_STRATEGY: ["AREA 0", "AREA 1", "AREA 2", ...]
```

This is just a rough sketch, but it doesn't look too complicated.
Even just wrapping the whole flow in a For can already be quite useful.

PR #795, opened 2025-10-14
feat: allow no filtering options in steps

KelvinChung2000:
When trying to drive flows within a flow, variable filtering becomes problematic because the config cannot see what is within the flow and will remove all variables.

### DIFF
--- a/librelane/steps/step.py
+++ b/librelane/steps/step.py
@@ -485,6 +485,7 @@ class Step(ABC):
         flow: Optional[Any] = None,
         _config_quiet: bool = False,
         _no_revalidate_conf: bool = False,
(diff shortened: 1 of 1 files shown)


donn:
Sorry for the hassle but I think I'd like to understand the use-case. By the sound of it, you have a top-level flow that incorporates one or more steps that themselves run another flow, which is quite unorthodox because we never exactly envisioned flows being used from inside steps.

This IS a low-risk PR and I'm willing to merge it because your `_` flag would [not become part of the API](https://librelane.readthedocs.io/en/latest/contributors/classes.html#internal) and if it causes issues we can revert it later. But I'd like to understand what you're doing because there may be a better way of doing it.

PR #811, opened 2025-11-11
fix: `expr::` would perform addition instead of subtraction

mole99:
That one was funny 😄

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,13 @@
 ## Documentation
 -->
 
+# 2.4.6
+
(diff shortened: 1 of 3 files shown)


donn:
wow im a complete idiot.

can you rebase this on `main`? this is worth a patch release. ill pull `main` into `dev` after

mole99:
It was an easy oversight 😁
But I really do appreciate the expression engine. It's absolutely great to be able to place macros based on the `DIE_AREA`.

Done! I hope the version bump is correct.

PR #819, opened 2025-11-21
feat!: enable DRT antenna repair by default

mole99:
This PR enables antenna repair after DRT by default. If I remember right, there were issues with rip-up and partial routing before? However, using recent OpenROAD versions this seems to work fine.

Let's see if we can resolve all antenna violations in the CI.

Additionally, the naming of the antenna repair variables are unified.

Note: One design in the CI (`regs_analog_ctrl`) required `allow_congestion` in GRT to be set.
The reason is that `repair_antennas` which runs after DRT uses GRT. However, the GRT options such as `allow_congestion` can only be set through calling `global_route` and not through `repair_antennas`.

The solution/workaround is to call GRT before perfoming the `repair_antennas` iterations.

---

TBD:

- Make antenna violations a flow failure just like for DRC and LVS.
- I would like to improve the manufacturability summary, e.g.:

|         | OpenROAD | KLayout | magic | netgen |
|---------|----------|---------|-------|--------|
| DRC     | ✅        | ✅       | ❌     |        |
| Antenna | ✅        | ❌       | ❌     |        |
| Density     |          | ❌       |  ✅     |       |
| LVS     |          | ❌       |       | ✅      |

It would be nice if the steps could add themselves to the summary?
This summary being in Markdown allows it to be added and rendered in the CI. Right now, this looks like [this.](https://github.com/wafer-space/gf180mcu-project-template/actions/runs/19543720582)

### DIFF
--- a/test/steps/xfails
+++ b/test/steps/xfails
@@ -1,0 +1,1 @@
+openroad.detailedrouting/004-fail-range-max # temporary


donn:
Sounds good. I went ahead and merged this. Not 100% sure if I want to flow error on all antenna issues bec I don't expect DRT antenna repair to fix non-trivial designs (the fastest test set is a very shallow pool)

Essenceia:
> Great to hear 🥳 Thanks for testing this, Julia!

My pleasure. 
Thank you for saving us half a tile :)

PR #832, opened 2025-12-12
chore: update nix-eda to 5.13.0

mole99:
Depends on https://github.com/librelane/librelane-ci-designs/pull/47


donn:
Blocked on https://github.com/IHP-GmbH/IHP-Open-PDK/pull/724 - IHP PDK update is necessary for new KLayout version but newer version has bugs with Magic DRC

PR #845, opened 2026-01-20
fix: swap ihp sealring width/height

mole99:
The first issue encountered with the padring generation! I hadn't noticed it before because I had always created square dies.

- The sealring PCell uses width/length, where width is along the y-axis and length is along the x-axis.
- The wrapper script inside the PDK, however, assigns width to width and height to length, which swaps the meaning.

To counteract this, we swap the width and height in `run_ihp_sg13g2`.


donn:
@mole99 I guess on this topic: I am curious as to how this works with relative floorplans?

Merging this because it's a bugfix, but we may need another approach to get this data…

donn:
@mole99 I specifically mean when `FP_SIZING` is not `absolute`

PR #851, opened 2026-02-06
fix: improve `from_magic_feedback` parsing

mole99:
This PR adds "-" to the wordchars of shlex, so that a string like "box 369787 -1 369789 1" will be split into:

`['box', '369787', '-1', '369789', '1']`

and not:

`['box', '369787', '-', '1', '369789', '1']`


mole99:
@donn Thanks!

PR #860, opened 2026-02-18
fix: load state with DesignFormat that is a list

mole99:
Currently, a DesignFormat can be a single file, or a dict (i.e. spef). However, in my plugin I would like to have a list of files as a DesignFormat. This already works partially: saving a state with a DesignFormat that is a list works. But loading the state again does not work.

This PR allows to load a state in which a DesignFormat is a list.

(This also fixes an unrelated issue where the wrong name for DiodeInsertion was displayed.)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,12 @@
 ## Documentation
 -->
 
+# 2.4.13
+
+## Misc. Enhancements/Bugfixes
+
+* Fix loading states if a DesignFormat is a list.
+
 # 2.4.12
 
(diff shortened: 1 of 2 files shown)


donn:
This is a bugfix and should be against `main`. You'll notice StateElement indicates that Lists should be supported already:

https://github.com/librelane/librelane/blob/8be0b6d15c62dbf964b3787704291cd2543c42ab/librelane/state/state.py#L43

So you're strictly fixing something I screwed up 🥲

PR #861, opened 2026-02-19
magic: use readspice before extraction

mole99:
To match the port order of the spice subckts when extracting from the layout we need to first read the spice models.


mole99:
Superseded by #863

PR #874, opened 2026-03-03
Corrected the procedures for the macro hardening guide

Jef-frey:
In the steps for macro hardening, there are some necessary scripts and files in 'openlane' directory but not in 'librelane' directory, this pull request provides the instructions on how to copy those files and update them to work in the 'librelane' directory. If there are ways to update and add said files to the 'librelane' directory directly, it should be done and these instructions can be ignored.

### DIFF
--- a/docs/source/usage/caravel/index.md
+++ b/docs/source/usage/caravel/index.md
@@ -140,6 +140,13 @@ ______________________________________________________________________
    $ mkdir -p ~/caravel_aes_accelerator/librelane/aes_wb_wrapper
    ```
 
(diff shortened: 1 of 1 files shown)


Jef-frey:
And I am not sure if this is the correct place to put this, but in the section [Re-running the flow with a modified configuration](https://librelane.readthedocs.io/en/stable/usage/caravel/index.html#id10), all the instructions about reconfiguring and rerunning aes_wb_wrapper feels moot. In the instruction the user_project_wrapper uses the aes_wb_wrapper as a macro, and after causing violations by using the old aes_wb_wrapper macro, a new aes_wb_wrapper macro is implemented and put in a different location in user_project_wrapper, however, since the physical view of the new macro is never saved/copied, the old macro is used and the only thing that changed and fixed the violations is the location of the macro.

donn:
I went ahead and rewrote the guide completely in #900 - thank you for pointing out the issues

PR #875, opened 2026-03-03
fix: add OpenSTA cell delay patch

mole99:
* Added a patch for fixing cell delays in certain situations

@urish Please give it a try.

### DIFF
--- a/nix/opensta.nix
+++ b/nix/opensta.nix
@@ -37,6 +37,10 @@ clangStdenv.mkDerivation (finalAttrs: {
     inherit sha256;
   };
 
+  patches = [
+    ./patches/opensta/fix_cell_delays.patch
+  ];
+
   postPatch = ''
     # utter bazel nonsense
     rm -f BUILD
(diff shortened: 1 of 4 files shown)


nix/opensta.nix
```diff
@@ -36,6 +36,10 @@ clangStdenv.mkDerivation (finalAttrs: {
     inherit rev;
     inherit sha256;
   };
+
```
donn: nit, should be no spaces on this line -- just run `nix fmt` when you add the magic patch tomorrow

donn (approved):

nix/opensta.nix
mole99: Whoops! Fixed. We should probably run nixfmt in the CI so that it can shout at me 😄

urish:
@mole99 LGTM

sorry for the delayed response, we're affected by the ongoing war

mole99:
No worries! I wanted to add it to the CI anyway.

PR #880, opened 2026-03-07
fix: reproducibles not working, drt snapshots

donn:
- update reproducible `@cloup_flow_opts` to not accept pdk options as they are unused
- openroad steps no longer run with the step directory as their cwd
  - was only required because earlier versions of openroad didn't support a drt snapshot directory
  - breaks reproducibles
- fixed botched DRT snapshot directory use in #702


mole99 (approved):
I wasn't aware that the cwd isn't supposed to be the step directory.

Thank you very much for fixing the OpenROAD reproducibles! This will make filing issues with OpenROAD much more pleasant.

PR #881, opened 2026-03-07
chore: opensta 3 compatibility (without updating tool)

donn:
- update io.tcl to create lln::get_corner_names, lln::get_corner_dict, and lln::set_sta_cmd_corner that abstract away whether opensta 2 or 3 is being used (or a corresponding version of openroad)
- rewrite dump_rc.tcl, set_rc.tcl, and corner.tcl to use said new functions

---

It's my hope that longer term, the `lln` Tcl namespace can include intuitive functions that users' custom OpenROAD steps can utilize. This is a good time to start it as any.

### DIFF
--- a/librelane/scripts/openroad/common/io.tcl
+++ b/librelane/scripts/openroad/common/io.tcl
@@ -1,4 +1,4 @@
-# Copyright 2025 LibreLane Contributors
+# Copyright 2025-2026 LibreLane Contributors
 #
 # Adapted from OpenLane
 #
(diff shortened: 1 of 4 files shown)


mole99 (approved):
As discussed, this is a great change, especially for when IIC-OSIC-TOOLS want to upgrade the OpenSTA version.

My only issue is the change to a "2025-2026" copyright date range. For legal reasons, this is [not required](https://aboutcode.org/2023/update-copyright-each-new-year/). Personally, I would prefer to keep the copyright date as the creation date. Updating this feels like unnecessary extra work to me.

donn:
@mole99 I didn't know that. I am okay not updating years moving forward and will update the PR to this effect.

mole99:
Thanks a lot, much appreciated!

PR #882, opened 2026-03-08
feat!: remove "readable_paths" from preprocessor

donn:
This makes refg:: (and by extension dir:: and pdk_dir::) work even outside the design and PDK directory.

This "feature" was initially done because remote execution was earlier in the roadmap back when the project was OpenLane 2 and it was thought this would enhance security by not allowing users to glob arbitrary directories in remote machines. But it's annoying and confusing users: we'll cross the bridge of securing LibreLane when we get to it. At the moment the Tcl-based files are a much easier point of attack anyway.

(I also removed a redundant antenna check in grt.tcl because it takes time for some designs and is repeated immediately after the next step.)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -543,6 +543,9 @@ Style Notes
 
   * `pdk_compat` set some compatibility values only if necessary.
 
+  * `refg::`, `dir::`, `pdk_dir::` now support globs outside the PDK directory
+    and the config directory.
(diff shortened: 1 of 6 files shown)


mole99 (approved):
Tested on ihp-sg13g2-librelane-template, works as advertised :)

PR #895, opened 2026-03-19
hotfix: skip StepException on empty SPEF dictionaries

donn:
When PDKs don't have RCX calibration files yet, an empty SPEF dictionary would cause a StepException to get raised.


mole99 (approved):
Looks good to me!

Issue #893, opened 2026-03-18
Macro grid autoplacer


mlyoung101:
### Description

This ticket is to track the macro grid autoplacer I suggested in the fossi-chat Matrix. Essentially, it would be an Python ODB step that would allow you to automatically place macros of the same type in a grid pattern.

I've started work over here: https://github.com/Precision-Contribution-Group/librelane/tree/grid-macro-autoplacer

### Proposal

This is my current proposal (ripped from docs I'm writing).

---

Sometimes, one would like to place a number of macros of the same type in a grid pattern. This is particularly
useful for SRAM arrays, but can also be useful for other patterns like FPGAs or some types of mixed-signal
designs.

Manually placing these macros in a grid is quite laborious. Luckily, there is now a step named
`Odb.MacroGridAutoplacer` that can help you with this.

To use this step, a grid must be declared as part of the macro instance using the **macro_grid** attribute.
For example:

```yaml
gf180mcu_fd_ip_sram__sram512x8m8wm1:
gds:
  - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/gds/gf180mcu_fd_ip_sram__sram512x8m8wm1.gds
lef:
  - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/lef/gf180mcu_fd_ip_sram__sram512x8m8wm1.lef
vh:
  - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/verilog/gf180mcu_fd_ip_sram__sram512x8m8wm1__blackbox.v
lib:
  - # removed for brevity
macro_grid:
    coord_bl: [5, 5]  # Coordinates (X, Y) of grid bottom-left corner in micrometres.
    size: 100         # Dimensions of grid (width X height) in micrometres.
    inst_pad_x: 2     # Padding between individual grid instances on the X axis in micrometres.
    inst_pad_y: 2     # Padding between grid rows (on the Y axis) in micrometres.
instances:
    ram1:
    ram2:
    ram3:
```

In the above example, three instances of GF180MCU SRAM will be attached to a macro grid.

---

The updated variable type looks like this:

```python
@dataclass
class MacroGrid:
    """
    A grid of macros for use in the grid macro autoplacer (see Odb.GridMacroAutoplacer).

    :param coord_bl: Coordinates (X, Y) of grid bottom-left corner in micrometres.
    :param size: Dimensions of grid (width X height) in micrometres.
[...]

PR #896, opened 2026-03-20
feat: Macro array instantiator

mlyoung101:
This PR adds a system to the preprocessor to instantiate an array of macros, as was described in #893.

It follows Leo's implementation description almost verbatim, with a few small deviations:
- I use `dimensions` not `size`, to try and make it more clear we are referring to rows X cols, not the size of the array
- I use a different syntax to perform inline variable substitutions, more on that below.

In order to implement the feature as Leo described, it's _also_ necessary to support inline variable substitutions in the preprocessor. Previously, the preprocessor would only let you reference a variable if the entire string was prefixed with `ref::`, and then only one variable per string. Now, you can also do `{VAR_NAME}` to reference a variable.

The macro array expansion system is implemented entirely in the preprocessor, there is are no additional passes or anything like that.

Please consider this PR an initial implementation, all feedback is welcome. The PR includes a number of unit tests to test both the macro expander, and the inline variable substitution system. I also updated the docs too. However, I haven't yet added a full end-to-end design test for the macro array expander, but I do intend to add this soon.


mlyoung101:
cc @mole99, @KelvinChung2000, @donn

mlyoung101:
I've now used it in a working tapeout, over here: https://github.com/SiliconPlatformsLab/wedgetail-v2

The config I made looks like this:

```yaml
  gf180mcu_fd_ip_sram__sram512x8m8wm1:
    gds:
      - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/gds/gf180mcu_fd_ip_sram__sram512x8m8wm1.gds
    lef:
      - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/lef/gf180mcu_fd_ip_sram__sram512x8m8wm1.lef
    vh:
      - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/verilog/gf180mcu_fd_ip_sram__sram512x8m8wm1__blackbox.v
    lib:
      "*_tt_025C_5v00":
        - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/lib/gf180mcu_fd_ip_sram__sram512x8m8wm1__tt_025C_5v00.lib
      "*_ff_n40C_5v50":
        - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/lib/gf180mcu_fd_ip_sram__sram512x8m8wm1__ff_n40C_5v50.lib
      "*_ss_125C_4v50":
        - pdk_dir::libs.ref/gf180mcu_fd_ip_sram/lib/gf180mcu_fd_ip_sram__sram512x8m8wm1__ss_125C_4v50.lib
    instances:
        # Make sure to specify the SRAMs in pdn_cfg.tcl
        "i_chip_core.detectors[{SEQ}].detector.sram":
            orientation: N
            array:
                offset: [492, 492]
                step: [442, 495] # SRAM size + 10 um padding
                dimensions: [3, 2]
```

And the result, after tapeout, looks like this:

image

I believe it's functioning exactly as intended.

This design is relatively simple and I could make an even simpler version to use as a test case.

KelvinChung2000 (commented):
A few minor things.  For YAML substitution, I would prefer `"{{ X }}"` or `"{{ .X }}"`, over `"{X}"`, as this is the more common templating syntax.

Maybe add a check that the `location` key cannot coexist with the `array` key, or else having `offset` and `location` might cause confusion.  Using the `location` to replace the `offset` might be an option as well. 

As a Python API user, I would prefer to call a function that returns the expanded `Instance` dict rather than passing in an `InstanceArray`. 

Since right now, I might need to write the code like the following
```python
m = Macro(
    gds="/foo",
    lef="/foo",
    instances={
        "inst_{X}_{Y}": Instance(
            location=None,
            orientation="N",
            array=InstanceArray(
                offset=(100, 100), step=(100, 100), dimensions=(2, 2)
            )),
    }
)
```
then

```python
m = Macro(
    gds="/foo",
    lef="/foo",
    instances=get_inst_array("inst_{X}_{Y}", offset=(100, 100), step=(100, 100))
)
``` 
Just to take advantage of the array expansion.

mlyoung101:
Thanks, Kelvin.

re: the YAML substitution, I agree, it's just like this at the moment because I can feed it through Python's `.format()` directly and I don't think you can format strings like `{{ VAR_NAME }}` by default in Python? My opinion is it's time we bring in a proper formatting engine like Jinja2, but that would probably need to be in another PR. 

I guess it depends how soon the other maintainers want to make a LibreLane v3 release, whatever makes it into that release should probably stay as the method for formatting strings.

re: the location/array check, that's a great idea, I'll add it.

re: the Python API, I agree, it's pretty cumbersome to use at the moment and I noticed this when writing the tests. Currently it's designed to fit into the existing config API as best as I could get it, _but_, I wouldn't be opposed to adding a `get_inst_array` routine as a separate function for API users that calls through the preprocessor, if that would suit your needs. I feel like that would be acceptable to add.

KelvinChung2000:
The `{{ }}` syntax will need a custom pre-processor. Probably eventually migrate and simplify to something like `${{ }}` for env substitution and `{{ }}` for variables. This will likely be in the next development cycle. 

What I would do is extract the expansion function and have the preprocessor call it; the call argument is populated by the YAML file. But if the intention is mainly to use it as a syntax for YAML pre-processing, I guess it does not matter. Since constructing such a grid is relatively simple when a user already touches Python/Librelane.

mlyoung101:
I did just find this: https://docs.python.org/3/library/string.html#template-strings-strings, maybe I could rewrite it using that if it's a good idea? It would support $VARNAME and ${VARNAME} syntax.

KelvinChung2000:
I think keeping it as is for now is fine.

librelane/config/preprocessor.py
```diff
@@ -230,6 +238,9 @@ def process_string(
     elif value.startswith(PDK_DIR_PREFIX):
         mutable = value.replace(PDK_DIR_PREFIX, f"refg::${Keys.pdkpath}/")
 
+    # before we do anything else, apply inline variable substitution
+    mutable = mutable.format(**symbols)
```
donn: This should probably require a `.meta.version` increment because current strings with {} in them would crash. To be fair I don't think there are too many of these but this behavior should be behind a flag, and `Config` should check if meta.version ≥ 3 (currently Tcl/JSON files default to .meta.version == 1 and YAML files default to .meta.version == 2) to enable this behavior.

I like this idea though.

librelane/config/preprocessor.py
```diff
@@ -388,13 +399,120 @@ def process_dict_recursive(
...
+    config_in: Mapping[str, Any],
+    exposed_variables: Dict[str, Any],
+) -> Mapping[str, Any]:
+    """
+    Looks for, and expands, macro arrays.
+
+    Returns the config with macro arrays fully expanded.
+    """
+    out = config_in
+    if (macros := config_in.get("MACROS")) is not None:
```
donn: Given this whole preprocessor is a hack and full of hacks I probably shouldn't be complaining about this, but I was hoping to move in the direction of less variable or type-specific hacks.

I would move this expansion to a `__post_init__` for `class Instance`:

https://github.com/librelane/librelane/blob/69b2067bd2b5eb89b84649b76e9edaa9e51e6735/librelane/config/variable.py#L89

---

The fun part is, to work with your other change, in .meta.version == 3 the instance names should be double escaped, i.e. `sram_instance_{{X}}_{{Y}}`…

donn (changes requested):
Some suggestions from my side, would appreciate if @mole99 could take a look though.

librelane/config/preprocessor.py
mlyoung101: Cool cool, that makes sense and good point. The other option we could do is maybe have a special case to handle '{}' strings directly? The reason being is that if we bump the version, I'd also like to include a proper formatting engine like Jinja2. Then again, we could bump the version to v3 now, and then I'll come back later and include Jinja2 and we can bump that to v4. What do you think?

mole99 (changes requested):
This PR is well done, thank you @mlyoung101 

I have one suggestion, in the diagram for macro instantiation, please start counting from 0 to match `SEQ`.
At first, I thought `SEQ` would start counting from 1, which thankfully it does not, and this would make it clearer.

librelane/config/preprocessor.py
mole99: I'm against integrating Jinja2 into LibreLane.

Jinja2 is very powerful, I've used it for website generation before. However, once you start using for loops and if clauses in a LibreLane configuration file, I believe it is better to drive LibreLane from a simple Python script. This is actually one of the strengths of LibreLane.

Initially, I thought we should use variable replacement with `{}` only for array instance names. However, I can see that a simple variable substitution can be generally useful in LibreLane configuration files without complicating things too much.

mole99:
One can definitely have much fun with this feature:

Bildschirmfoto vom 2026-03-30 15-07-24

librelane/config/preprocessor.py
mlyoung101: Gone ahead and implemented this, it *seems* to work, still checking on it. Funnily enough, the double expansion `{{}}` isn't necessary, because apparently the preprocessor never checks the _keys_ of the macro fields, they are not expanded. This is probably (?) a bug, but I'm leaving it in place for now. I assume we should probably fix this.

Just need to add the version bump (I'm writing this so I remember to do it).

Finally, I actually had to put it as the `__post_init__` for Macro, because Instance would need to access its parent otherwise (to add the sub-instances) which is not currently possible.

mlyoung101: image

yeah, looks like it's still behaving the same in my design (same YAML and everything, just changed the LibreLane branch)

donn:
@mole99 to take one last look…

mole99 (changes requested):
I have now also tested the substitution and it looks fine to me.

The one thing missing is gating it behind meta.version ≥ 3 and bumping .meta.version for yaml files to 3.
Please also rebase on top of dev, as now the proper 3.1.0 section was added to the changelog.

Then we're good to go! 🤞

mlyoung101:
Cool; @mole99 @donn:

- Merged dev into the branch
- Gated behind meta.version==3 (the logic is kind of ugly since I have to plumb the version into the preprocessor, sorry about that, I don't think there's a better way)
- Bumped default meta version to 3 for YAML documents
- Moved notes around in changelog

Since the PR was last touched, I've also changed my preferred name; I took the liberty of updating it in Authors.md. Hope that's okay!

mole99 (approved):
Thank you for your contribution, Mel!

librelane/config/config.py
```diff
@@ -353,7 +353,7 @@ def get_meta(
         :param config_in: A configuration object or file.
         :returns: Either a Meta object, or if the file is invalid, None.
         """
-        default_meta_version = 2
+        default_meta_version = 3
```
donn: Sorry Mel, this would still be an API break lol. Default should remain 2 or else existing scripts with `{}` in them might break. But other than that this is ready to merge.

donn:
I'm just gonna merge it so I can figure out wth's wrong with the CI at my own pace.

Issue #901, opened 2026-03-25
Read flist (*.f) files


mlyoung101:
### Description

LibreLane should be able to read *.f files that are commonly used in industry and with tools, e.g. [Bender](https://github.com/pulp-platform/bender). They look like this:

```
+incdir+/home/matt/workspace/currawong/.bender/git/checkouts/axi-93c4dff6ad7bb0cf/include
+incdir+/home/matt/workspace/currawong/.bender/git/checkouts/common_cells-b5d7706e88172c77/include
+incdir+/home/matt/workspace/currawong/.bender/git/checkouts/ibex-7c8e10aa13f0ed7c/rtl
+incdir+/home/matt/workspace/currawong/.bender/git/checkouts/ibex-7c8e10aa13f0ed7c/vendor/lowrisc_ip/dv/sv/dv_utils
+incdir+/home/matt/workspace/currawong/.bender/git/checkouts/ibex-7c8e10aa13f0ed7c/vendor/lowrisc_ip/ip/prim/rtl
+define+TARGET_FLIST
+define+TARGET_SYNTHESIS
/home/matt/workspace/currawong/.bender/git/checkouts/tech_cells_generic-72fdc0c045a93f86/src/rtl/tc_sram.sv
/home/matt/workspace/currawong/.bender/git/checkouts/tech_cells_generic-72fdc0c045a93f86/src/rtl/tc_sram_impl.sv
/home/matt/workspace/currawong/.bender/git/checkouts/tech_cells_generic-72fdc0c045a93f86/src/rtl/tc_clk.sv
```

### Proposal

My opinion is that we should introduce a new optional config variable, `VERILOG_FLIST_FILES`. We only support (System)Verilog F-list files for now.

Inside the preprocessor, before the main config parse is run, we check if a config variable called `VERILOG_FLIST_FILES` exists as a list.

If it does, we go thru each flist file, and parse each directive. We either add it to `VERILOG_INCLUDE_DIRS`, `VERILOG_DEFINES`, or `VERILOG_FILES` respectively.

Then, we continue on as normal.

Flist files should support being specified like `dir::../foo.f`.

This addition will be documented in the changelog and also in the Pre-processing section of the user manual.

mlyoung101:
cc @mole99

PR #902, opened 2026-03-25
feat: Add f-list (*.f) parser

mlyoung101:
This adds a F-list/file list/*.f file parser through a new config key, `VERILOG_FLIST_FILES`. It's implemented as part of the pre-processing stage, so no downstream tasks are aware of its existence. The pre-processor will locate the *.f files, parse them, and append the resulting files to `VERILOG_FILES`, `VERILOG_DEFINES` or `VERILOG_INCLUDE_DIRS` as needed. Globbing and other preprocessor directives are supported as part of the `VERILOG_FLIST_FILES` key, but are not supported in the actual F-list file itself. VHDL is currently unsupported.

This has been implemented verbatim through the design described by myself in #901, which in turn was earlier posted on the fossi-chat Matrix.

Docs updates and tests are included as well :cat:


donn:
Took me a while to get to this but I believe slang already has native support for .f files, why involve the preprocessor at all?

mlyoung101:
Slang does indeed have native support for F-lists, but I wrote this because if you pass a .f file into `VERILOG_FILES` it'll be interpreted by LibreLane as a Verilog file not as an f-list. If we can get access to pyslang though inside LibreLane, we could resolve the F-list through them rather than my Python hackery

ricted98:
Hi there! I think this feature would come in handy, since flist files are quite common. Indeed I ended up here looking for such support in LibreLane. Is there any ongoing discussion on how to get this feature upstreamed? For example, would exposing and using the flist parser of the slang frontend be an acceptable solution, as suggested by @donn? Thanks!

donn:
Sorry for the wait, should be bumping LibreLane to Yosys 0.68 soonish which just incorporates slang. We can then just use that frontend.

mlyoung101:
That makes sense yea; when that drops I'll rewrite the `USE_SLANG` stuff to use the built-in Slang. Did you want me to close this one out then donn?

ricted98:
Thanks for the update! I assume @mlyoung101 already has a solution in mind, but let me know if you need an extra set of hands. I recently started exploring LibreLane and would be more than happy to help upstream this feature to get used with the codebase.

mlyoung101:
Sure, absolutely, it's great to have new contributors on board so I would happy to leave this one up to you @ricted98, if you'd like to get acquainted with the codebase.

The general procedure once Slang drops in Yosys upstream would just be to cease loading it as a plugin in pyosys.py (I believe) and use the then built-in commands. Should be only a couple of lines change.

My personal preference would also be to make USE_SLANG the default as its objectively better in every aspect IMHO compared to the old Bison parser. However there is a possibility that we break some old designs with obscure parse behaviour or whatever.

PR #905, opened 2026-03-26
Add gdsfill as filler

wallento:
Run gdsfill as filler tool. Set the following substituting step:

    KLayout.Filler: Gdsfill.Filler


donn:
@wallento Why did you close this?

PR #906, opened 2026-03-26
Add gdsfill as filler

wallento:
Run gdsfill as filler tool. Set the following substituting step:

    KLayout.Filler: Gdsfill.Filler


wallento:
I had to add some mocking into doc for the import of the external module

donn:
It took me so long to review this that the Gdsfill developers opened their own version of the PR… #984 

That approach looks more straightfwd, so let's just use that.

PR #909, opened 2026-04-02
bugfix: backport The-OpenROAD-Project/OpenROAD#9975

donn:
See https://github.com/The-OpenROAD-Project/OpenROAD/discussions/9814.

Also added a number of missing Qt plugins since we're rebuilding OpenROAD anyway so might as well.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,14 @@ Style Notes
 
 -->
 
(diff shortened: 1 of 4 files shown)


mole99 (approved):
The backport looks fine to me. Thanks also for adding the missing plugins!

PR #919, opened 2026-04-13
chore: move to using ruff only

KelvinChung2000:
Ruff itself should be able to replace both black and flake8.

If you would like to enable more rules, please let me know.


donn:
Not opposed to the change because flake8 is annoying but you really should have asked in an issue first lol

@mole99 Light reading https://docs.astral.sh/ruff/formatter/black/

KelvinChung2000:
I think this is just, in general, the right thing to do. I got annoyed by this kind of formatting problem, and the current setup makes absolutely no sense.  This only took me 10 min to do so, so I am not that upset if this is not accepted.

donn:
It’s more of a process than a technical issue, technically this makes all the sense. Practically we’d like to align changes like that with releases or whatnot.

It’s not a huge issue but just for future reference.

mole99:
Looks fine to me.

I don't generally care which linter or formatting is used, as long as one is used.
Not really happy about who plans to acquire Astral, but I guess if that ever becomes an issue, we can switch back.

However, this should probably target dev, where active development is happening?

KelvinChung2000:
I got the impression somewhere that `dev` vs `main` won't be used anymore.  But anyway, once this is landed on `dev`, I think you might want to cherry-pick it to main as well, since otherwise the re-merge will have a lot of linting noise. 

Do you want to enable more rules, such as docs, import sort and annotation checking?  We could do this in one go, or we can have multiple PR, with each PR enabling one extra rule.

donn (approved):
LGTM, though we're going to merge it closer to 3.1's release bec if we merge it now we have to ask everyone to reformat their PRs and stuff.

PR #921, opened 2026-04-18
feat: make verilator error on multiply-driven nets

Ravenslofty:
A user came to me for help with a design where Yosys produced a lot of check warnings due to multiply-driven nets. Looking through the run logs, Verilator caught the conflicts too, but only issued a warning which got lost in the backlog.

I think this warning should be upgraded to an error because Verilator can provide much more useful reporting with source locations compared to Yosys.

### DIFF
--- a/librelane/steps/verilator.py
+++ b/librelane/steps/verilator.py
@@ -78,6 +78,12 @@ class Lint(Step):
             default=True,
         ),
         Variable(
+            "LINTER_ERROR_ON_MULTIDRIVEN",
+            bool,
+            "When a net has multiple drivers, report this as a linter error.",
+            default=True,
+        ),
+        Variable(
             "VERILOG_DEFINES",
             Optional[List[str]],
             "Preprocessor defines for input Verilog files",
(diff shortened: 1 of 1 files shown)


Ravenslofty:
um, the CI failures look transient?

librelane/steps/verilator.py
```diff
@@ -206,6 +206,9 @@ def run(self, state_in: State, **kwargs) -> Tuple[ViewsUpdate, MetricsUpdate]:
         if linter_vlt := self.config["LINTER_VLT"]:
             extra_args.append(linter_vlt)
 
+        # It's more user-friendly to catch multiple-driver conflicts here in Verilator (if possible) than later in Yosys.
+        extra_args.append("--Werror-MULTIDRIVEN")
```
donn: Should be added before LINTER_VLT I think so options from LINTER_VLT could override it (assuming that's how Verilator handles that?)

Other than that LGTM. @mole99 Though unorthodox, I think we should let this be a patch release? It catches bad designs early and is fairly low-impact otherwise

donn:
CI issues are transient yes, the Nix installation action is a bit flaky. pun intended

librelane/steps/verilator.py
mole99: Yes, and I would move this option behind a `LINTER_ERROR_ON_MULTIDRIVEN` variable (enabled by default), underneath `LINTER_ERROR_ON_LATCH` a few lines above.

Fine by me to merge into main.

mole99 (approved):
LGTM thanks!

PR #925, opened 2026-04-20
feat: add support for sky130 padring generation

mole99:
This PR enables padring generation on sky130, together with https://github.com/RTimothyEdwards/open_pdks/pull/506.

## Steps

* OpenROAD.PadRing

  * Added `PAD_ROTATION_[HORIZONTAL|VERTICAL|CORNER]` for pad cells that require it (e.g. sky130).

* OpenROAD.GeneratePDN

  * Added `PDN_CORE_RING_CONNECT_TO_PAD_LAYERS` to restrict the layers for connecting to pads.

## Misc. Enhancements/Bugfixes

- Split `LIB` into `CELL_LIBS` and `PAD_LIBS`.


mole99:
Ready to merge! PDK update and CI full-chip test design will come later.

mole99:
So I was wondering why updating the PDK breaks the CI.

Turns out TIMING_VIOLATION_CORNERS was `"*tt*"` and is now `"*"`. I agree with the current setting, the corners in sky130 are reasonable and should all be checked. Users can change it back to `"*tt*"` if they want to.

I'll update the CI designs to reduce the target frequency.

donn:
Yes, @kareefardi and I decided that was a necessary change for then OpenLane 3.0. *tt* was never good enough but * was a breaking change. Kinda.

mole99:
@donn Now it's really ready :)

librelane/config/flow.py
```diff
@@ -535,6 +537,27 @@ def _prefix_to_wildcard(prefixes_raw: Union[str, Sequence[str]]):
         units="µm",
         pdk=True,
     ),
+    Variable(
+        "PAD_ROTATION_HORIZONTAL",
+        Optional[str],
```
donn: Shouldn't these be the Orientation enum?

https://github.com/librelane/librelane/blob/1e4f4d5bf9d2693798b12dc0c1cd0337ad266a0d/librelane/config/variable.py#L65

donn (commented):
Minor comment but LGTM

librelane/config/flow.py
mole99: Hm, not really. It seems to use different values such as R0 and MXR90: https://openroad.readthedocs.io/en/latest/main/src/pad/README.html#id5

Maybe we need a new enum for that too?

mole99: Found the mapping: https://github.com/The-OpenROAD-Project/OpenROAD/blob/d87cb90924369d20c1eac56a3a6db9241f9e8792/src/odb/src/db/dbTypes.cpp#L20

Can we have two values mapping to the same enum?

donn: We do have that in the Enum you'll notice, multiple Enums map to the same string value

mole99: Right, I'm blind 🤦‍♂️

Changing it too `Orientation`...

mole99: The problem with changing it to `Orientation` is that I then get values such as "N", "FN", "W" etc. while OpenROAD expects R0, MY, R90 etc.

Any ideas?

donn: OpenROAD should handle both, no? The mapping you linked me to does, at least

mole99: No, unfortunately `make_io_sites` throws an error when passing LEF orientations. We could add a second enum `ODBOrientation` if needed?

donn: Nah, that'd be confusing. Just add a `validator=` that checks its a valid key for now and starting 4.0.0 what we can do is change the values for Orientation to be the RX format

mole99: I used a `Literal` with all options. That should also be fine?

donn: better, actually

donn (approved):
feel free to merge

PR #929, opened 2026-04-24
doc: referencing deprecated variable name FP_PDN_RAILS_LAYER

jeras:


donn (approved):
Thanks for the catch!

PR #936, opened 2026-05-06
ci: add sky130 full-chip design

mole99:
This adds a full-chip design to the CI.

I added a placement blockage in the center of the chip to speed up the flow:

grafik

### DIFF
--- a/.github/test_sets/test_sets.yml
+++ b/.github/test_sets/test_sets.yml
@@ -54,6 +54,9 @@
       script: harden.py
     - name: hold_violations_3
       script: harden.py
+    # Full chip design
+    - name: full_chip_sky130
+      config_file: config.yaml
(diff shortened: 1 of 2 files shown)


mole99:
This is ready to be merged 🙌

I currently ignore a few KLayout DRC violations, this is because OpenROAD ignores POLYGON statements in the LEF and therefore routes too close to some other structures.
One way to work around this is to place blockages, but I have decided to leave it in the CI like this until the issue is fixed in OpenROAD.

In the near future, I will probably rename full_chip to full_chip_ihp_sg13 and perhaps add one for gf180mcu as well.
It would, in theory, be possible to create a unified design for all three PDKs (anyone interested?), but the practical use would be limited.

mole99:
Merging this now since it ensures there are now future regressions with sky130 full chip designs!

PR #937, opened 2026-05-07
feat: add support for choosing OpenROAD thread count

mlyoung101:
Quick PR that adds support for selecting thread count in OpenROAD, should (hopefully) improve performance. 

Currently, we only support multi-threading in DRT, but other stages like GPL are increasingly supporting multi-threading now too. There is mixed evidence (https://github.com/The-OpenROAD-Project/OpenROAD/issues/7159, https://github.com/The-OpenROAD-Project/OpenROAD/issues/5626) about whether or not this actually helps though.

The way I plumb the new variable `OPENROAD_THREADS` into the `get_command()` routine is really ugly, but I couldn't think of a better way to do it. Another option would be to call the `set_num_threads` command in the Tcl; but we'd have to do that for every script since there's no common OR Tcl script. The way it is currently, we get support now and in the future if OR upstream adds any more multi-threading stuff without changing our code.

I did not add a unit test, but I manually verified that the OpenROAD command is prefixed with the correct thread count, e.g.:

image

I suggest that before this is merged, we deprecate `DRT_THREADS`, as this supersedes it.

Note also that by default, OpenROAD uses only 1 thread (!)


mole99:
Hi @mlyoung101, it seems that you don't use the `OPENROAD_THREADS` variable through `self.config` atm.

We shouldn't access `OPENROAD_THREADS` through `os.environ`. If needed, you should be able to set it externally via `librelane -c OPENROAD_THREADS=n`.

mlyoung101:
hey Leo! Yes sorry I believe the way I was trying to do this at the time is a bit of a hack, I was trying to pass the variable through `os.environ`.

It looks like upstream is going to default to this behaviour though: https://github.com/The-OpenROAD-Project/OpenROAD/pull/10724

Do you reckon it's better we just wait for upstream to merge this, or do we want to enforce it on our side as well?

donn:
No, I like this being on our side, if you fix the config thing I will merge it

mlyoung101:
@donn lemme know what you think of this. I don't know what I was smoking with that `os.env` stuff lol, this is much easier. Seems to work based on the smoke test result.

librelane/steps/openroad.py
```diff
@@ -291,6 +291,12 @@ class OpenROADStep(TclStep):
             "Cull duplicate IPVT corners during PNR, i.e. corners that share the same set of lib files and values for LAYERS_RC and VIAS_R as another corner are not considered outside of STA.",
             default=False,
         ),
+        Variable(
+            "OPENROAD_THREADS",
+            Optional[int],
+            "The number of threads OpenROAD may use. If unset, this will be equal to the machine's thread count by default.",
+            default=_get_process_limit(),
```
donn: Was gonna fix this myself but thought you'd be interested to learn this since you're a frequent contributor: it's better to just let the default be `None` because then the autogenerated documentation would be whatever the number of processors is on `readthedocs.io`'s builders

librelane/steps/openroad.py
```diff
@@ -509,9 +515,13 @@ def __hash__(self):
 
     def get_command(self) -> List[str]:
         metrics_path = os.path.join(self.step_dir, "or_metrics_out.json")
+        threads = str(self.config["OPENROAD_THREADS"])
```
donn:
```suggestion
        threads = str(self.config["OPENROAD_THREADS"] or _get_process_limit())
```

librelane/steps/openroad.py
```diff
@@ -509,9 +515,13 @@ def __hash__(self):
...
+        info(f"OpenROAD will use {threads} threads")
```
donn:
```suggestion
        verbose(f"Running OpenROAD with {threads} threads…")
```

librelane/steps/openroad.py
mlyoung101: ahhh okay yes I thought this might be the case!! good to know, ty ty

librelane/steps/openroad.py
mlyoung101: I have applied this, although personally I'd like to see this printed to the console at least once per session, however in fairness I am usually a fan of extremely verbose logging

mlyoung101:
I just saw this from my original message that I totally forgot about:

> **I suggest that before this is merged, we deprecate DRT_THREADS, as this supersedes it.**

Does that still seem reasonable? Should I go ahead and do that?

donn:
Oh, yes, if you can confirm DRT runs multi-threaded by default when `openroad -threads` is > 1 then I don't mind.

mlyoung101:
So based on some quick grepping:
- LibreLane sets `DRT_THREADS`
- That's read by LibreLane's `drt.tcl`, which calls `set_thread_count`
- OR defines `set_thread_count` in `OpenRoad.i` etc; and so it seems to apply to the whole process in fact.

Given this I'll go ahead and deprecate it and update the changelog too

donn:
Last thing is you should remove the reference to `DRT_THREADS` in `librelane/scripts/openroad/drt.tcl`.

donn:
…now add `openroad.detailedrouting/006-success-threads-single`  and `openroad.detailedrouting/007-success-threads-multiple` to `test/steps/xfails`…

(sorry for the crash course on librelane maintenance,)

mlyoung101:
haha no you're good no worries! (and sorry that this is stored under my org so you can't just edit it directly :sob:)

PR #938, opened 2026-05-07
ci: update deprecated github actions

jeras:
I am trying to run CI on my LibreLane fork and encountered a bunch of warning about using old versions of GitHub actions: `Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-python@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true.`

I updated the actions and this removed the warnings, while it does not seem to have caused any regressions. But I am not sure, whether CI passes or fails seems somehow un-deterministic.

This is the CI on my fork (for the commit in this pull request): https://github.com/jeras/librelane/actions/runs/25518401698

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       gf180mcu_hash: ${{ steps.set-rev.outputs.gf180mcu_hash }}
       ihp-sg13g2_hash: ${{ steps.set-rev.outputs.ihp-sg13g2_hash }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - name: Python Dependencies
         run: |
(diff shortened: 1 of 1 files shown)


jeras:
@donn could you look at this PR, it is trivial, and it removes a bunch of annoying warnings in the CI.

donn (approved):
Thank you

PR #943, opened 2026-05-12
fix: include EXTRA_LIBS and macro lib views in synthesis lib set

IAmMarcelJung:
By adding this, yosys can then also report the area for cells added in `EXTRA_LIBS`. Before I got the warning
`Area for cell type \cell_added_to_extra_libs is unknown!` This was previously fixed in OpenLane (The-OpenROAD-Project/OpenLane#1646), but I think it got lost in the rewrite.

### DIFF
--- a/librelane/steps/pyosys.py
+++ b/librelane/steps/pyosys.py
@@ -335,10 +335,13 @@ class VerilogStep(PyosysStep):
                 DesignFormat.LIB,
             ]
         )
-        for view, _ in self.toolbox.get_macro_views_by_priority(
+        macro_lib_views = []
+        for view, fmt in self.toolbox.get_macro_views_by_priority(
             self.config, format_list
         ):
             blackbox_models.append(str(view))
+            if fmt == DesignFormat.LIB:
+                macro_lib_views.append(str(view))
 
         if libs := self.config.get("EXTRA_LIBS"):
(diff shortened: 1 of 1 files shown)


donn:
!metrics

librelane/steps/pyosys.py
```diff
@@ -328,10 +335,13 @@ def get_command(self, state_in: State) -> List[str]:
                 DesignFormat.LIB,
             ]
         )
-        for view, _ in self.toolbox.get_macro_views_by_priority(
+        macro_lib_views = []
+        for view, fmt in self.toolbox.get_macro_views_by_priority(
```
donn, line 339: Liberty is the lowest priority here, so it would only get added if there are no other views.

donn (changes requested):
Sorry to keep you waiting!

My concern here is that for macros, historically at least, OpenSTA tends to write subpar liberty files. This especially counts for things like power pins and definitely timing information.

Your approach makes more sense, but I'm not sure if the blackbox/verilog header approach takes priority or the lib would take priority.

There is a method by which we can only use these lib files for calculating area, which is by reading all liberty files unconditionally but only passing them to area. Would that work for your purposes?

(We can do that ourselves, we've kept you waiting long enough.)

Issue #949, opened 2026-05-21
Remove all the $check Yosys cells.


nathanchandanson:
### Description

Currently, even though the Yosys script specifies `d.run_pass("hierarchy", "-check", "-top", top, "-nokeep_prints", "-nokeep_asserts")`, the `-nokeep_asserts` does not completely remove all the assertions.
This can lead to unmapped `$check` cells at the end of the Yosys.Synthesis step.

It could be nice to completely remove all the formal stuff as it is not needed for an ASIC flow.

### Proposal

I added the `d.run_pass("chformal", "-remove")` command just after the `d.run_pass("hierarchy", "-check", "-top", top, "-nokeep_prints", "-nokeep_asserts")` one.

<img width="943" height="124" alt="Image" src="https://github.com/user-attachments/assets/34a7e3dc-3eee-4a01-98dc-acf01361ce5b" />

I am not sure this is the best way to do it but it solved my problem.

donn:
That'd be perfect, actually. Could you open a PR for that?

nathanchandanson:
Yes no problem I will do it this afternoon.

nathanchandanson:
Implemented in #950.

PR #950, opened 2026-05-22
fix: remove all the formal cells during synthesis

nathanchandanson:
This PR removes the formal cells (`$check`, ...) during the synthesis pass as explained in #949.

This is done by running `chformal -remove` at the beginning of the synthesis script.


donn (approved):
Looks good to me, thank you so much for the contribution!

PR #953, opened 2026-05-26
cleanup: removed unused variables SYNTH_MUX_MAP/SYNTH_MUX4_MAP

jeras:
After investigating Yosys technology mapping files inside `open_pdks` `sky130` I noticed the `SYNTH_MUX_MAP`/`SYNTH_MUX4_MAP` variables and therefore the mapping files are not used. Nevertheless Yosys and ABC do a good job mapping multiplexers to standard cells.

The question remain whether to also remove related code and files from `open_pdks`.

I have removed the variables and updated the changelog.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -39,6 +39,12 @@ Style Notes
   * Upgraded warnings on multiply-driven nets to an error that can be disabled
     by setting `LINTER_ERROR_ON_MULTIDRIVEN` to `false`.
 
+* `Yosys.*Synthesis`
+
+  * Removed unused variables `SYNTH_MUX_MAP`/`SYNTH_MUX4_MAP`.
+    As a note, Yosys and ABC are able to map multiplexers to SCL cells
+    without explicit technology mapping.
+
 ## Misc. Enhancements/Bugfixes
 
 * `openlane.config`
--- a/librelane/steps/pyosys.py
+++ b/librelane/steps/pyosys.py
@@ -195,18 +195,6 @@ class PyosysStep(Step):
             pdk=True,
         ),
         Variable(
-            "SYNTH_MUX_MAP",
-            Optional[Path],
-            "A path to a file containing the mux mapping for Yosys.",
-            pdk=True,
-        ),
-        Variable(
-            "SYNTH_MUX4_MAP",
-            Optional[Path],
-            "A path to a file containing the mux4 mapping for Yosys.",
-            pdk=True,
-        ),
-        Variable(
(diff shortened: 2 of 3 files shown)


mole99 (approved):
LGTM a cleanup seems sensible to me to prevent confusion

donn:
I know this is pedantic but it's technically a breaking change from the documentation. I'm not opposed to twisting the rules a bit considering it literally never worked in LibreLane compared to OpenLane but also wouldn't making them work be an option?

jeras:
I think it might be possible to get MUX mapping rules to work, but it might yield worse results than just letting Yosys+ABC do it.

My main concern would be the added complexity to LibreLane and PDKs that would arguably provide no gains, while it would increase the required maintenance efforts for future versions of LibreLane and OpenROAD.

There is also the issue how the current MUX mapping rules create a perceived limitation in LibreLane capabilities. The rules map only `mux2` and `mux4` cells, while Yosys+ABC are also able to use `mux2i` cells. Somebody debugging multiplexer complexity in their design might spend unnecessary effort researching and writing their own custom MUX technology mapping rules instead of letting Yosys handle it. If there is some issue which only MUX mapping can solve the user would still be able to use the LibreLane `SYNTH_EXTRA_MAPPING_FILE` just for their project.

Actually `mux2i` cells (inverted output, the cell is smaller than `mux2`) were on the SYTH_EXCLUDED list in the PDK before I removed them from the list during the cleanup (thus allowing synthesis to use them). Unless Yosys+ABC has some problems mapping them this should generally translate into better multiplexer synthesis. I tested it on an admittedly trivial set of multiplexers of different 2**N sizes. While older versions of Yosys (0.62) used a mix of muxes and other cells, The latest release 0.65 constructed a mux tree from only mux cells. Further testing might show other issues, but this is not something MUX technology mapping is supposet or able to solve.

I will check again whether MUX mapping is mentioned elsewhere in the LibreLane documentation. This PR already remove the obvious parts and adds a clear discontinuation notice in `‎Changelog.md`.

mole99:
I agree that we should simply let Yosys+ABC do the mux mapping.
There does not seem to be a benefit in updating LibreLane to map them explicitly.
Removing them clarifies things and makes it a bit less effort to bring up a new PDK.

donn:
Fair enough. I'm willing to just call it a documentation bug, but in that scenario I would ask you to rebase this on `main`.

jeras:
@donn, I did the rebase of the change to `main`.

donn (approved):

PR #959, opened 2026-06-07
feat: magic: disable locking, crashbackups and drc where appropriate

### LINKED ISSUE #818: Check or set the maximum number of open file descriptors
### Description

Leading up to the first wafer.space shuttle, two users faced the issue that the max open file descriptors count was set too low by default. One user is on Ubuntu and the other on macOS, both are using Nix.

It seems that magic, in particular, needs to open a lot of file descriptors due to its hierarchy of mag files.

### Proposal

There are two approaches we could take to improve this situation:

1. Report a warning when the max open file descriptor count for the process is too low. (What is too low?)
2. Increase the soft limit, perhaps even up to the hard limit.

If there is no downside to 2, I would opt for it, since it doesn't require any user interaction.
If we go for 2, the question is when should ulimit be called?
Should this be done in LibreLane or as a shell hook?

Would appreciate your input @donn 👏

mole99:
Depending on the script we can disable crashbackups and locking, as well as drc when not needed.

Fixes #818

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,14 @@ Style Notes
 
 -->
 
+# 3.0.4
+
+## Steps
+
+* `Magic.*`
+
+  * Added `locking disable` to scripts to prevent exceeding max open file descriptor limit.
+
 # 3.0.3
 
 ## Steps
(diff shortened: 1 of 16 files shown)


donn:
@mole99 Shouldn't this be in main?

mole99:
@donn If it doesn't cause any issues, agreed :)
I'll retarget the PR.

donn (approved):

PR #964, opened 2026-06-15
feat: KLayout.Render: allow to run with GDS only

mole99:
I'm using this step in the gf180mcu precheck, which only provides a GDS.


mole99:
Merging this one as it's a minor change.

PR #965, opened 2026-06-16
Update padring generation

mole99:
* OpenROAD.PadRing

  * Added `PAD_SPACING_MULTIPLE` to make the pad spacing configurable.

  * Added `PAD_TRIM_ROWS` to allow for pad rings with fewer than four rows.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -54,6 +54,8 @@ Style Notes
 
   * Added `PAD_SPACING_MULTIPLE` to make the pad spacing configurable.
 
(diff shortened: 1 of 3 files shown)


mole99:
The first image shows that the pads for the gf180mcu 1x1 slot are centered again.
The second image shows the 0p5x0p5 slot with only one row and `PAD_TRIM_ROWS` enabled.

|  Bildschirmfoto vom 2026-06-16 09-18-33 | grafik  |
|---|---|

mole99:
Sylvain had a look at the output and confirmed that the fix works as expected, i.e. the pads are centered correctly.

PR #972, opened 2026-06-25
fix: make KLAYOUT_DRC_OPTIONS of type `Dict[str, str]`

mole99:


donn:
Hi Leo, this is a breaking change. Is this absolutely necessary? Why not just stringify them on demand?

mole99:
Hi @donn, how would I stringify them on demand? I was thinking of always keeping them as a string.

The problem is with casting integers and booleans, which can be ambiguous.

For example, on gf180mcu, we need to set:

```
KLAYOUT_DRC_OPTIONS:
  workers: 1
  ...
```

Which, in the case of `Optional[Dict[str, Union[bool, int, str]]],`, gets converted to `workers: true`. However, the KLayout DRC deck expects an integer, i.e. the number of workers.

Thinking I was being clever, I swapped them around to the current `Optional[Dict[str, Union[int, bool, str]]],` which fixes the issue for gf180mcu, however, ihp-sg13 sets the following options:

```
KLAYOUT_DRC_OPTIONS:
  no_recommended: true
  ...
```

This is parsed to `no_recommended: 1`. However, the KLayout DRC deck expects `false`/`true.`
So, as I see it, there's no way to win this game (except ensuring `true` does not get converted to `1` for `int`?).

However, we could skip all of that if we simply passed on the string as a string, since it gets supplied as an argument to KLayout anyway. This way there's no ambiguity involved.

donn:
I can work on a solution for that. More importantly though, this shouldn't be happening at all with meta.version ≥ 2 so I need to investigate what's happening exactly.

donn:
Superseded by #994

PR #973, opened 2026-06-27
KLayout.Render: fix missing optional inputs

donn:
- properly declare DEF and GDS as optional inputs
- do not abort the flow if both are missing (as they're technically optional)

---

I suppose a better way to handle this would be to develop a predicate system for inputs where it can be `DesignFormat.DEF | DesignFormat.GDS` but this would do for now

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -39,20 +39,21 @@ Style Notes
 
   * `run_sky130`: added `thr` to match DRC deck.
 
-* KLayout.Render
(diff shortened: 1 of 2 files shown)


mole99:
I had completely forgotten about `.mkOptional()`. Thanks Donn!
Not aborting the flow also makes sense.

mole99 (approved):

PR #982, opened 2026-07-14
chore: update gf180mcu PDK hash

mole99:
This fixes an issue in the gf180mcu_as_sc_mcu7t3v3 SCL.


mole99:
Merging this one since it's a critical fix of the SCL with no other effects.

PR #983, opened 2026-07-16
feat: support for multiples LINTER_VLT files

nathanchandanson:
This allows to specify a list of paths to `.vlt` files for the `LINTER_VLT` variable.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -29,7 +29,7 @@ Style Notes
 
 ## Steps
 
-* KLayout.DRC
+* `KLayout.DRC`
 
   * `KLAYOUT_DRC_OPTIONS`: changed order of type evaluation so that `threads: 1` is evaluated as integer.
 
@@ -65,7 +65,7 @@ Style Notes
 
   * Added `PDN_CORE_RING_CONNECT_TO_PAD_LAYERS` to restrict the layers for connecting to pads.
 
-* OpenROAD.RepairDesign
+* `OpenROAD.RepairDesign`
 
   * Use `SYNTH_BUFFER_CELL` for port buffering.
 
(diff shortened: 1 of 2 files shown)


mole99:
@nathanchandanson Thanks for the PR! Please also update the Changelog, as done in other PRs.

nathanchandanson:
@mole99 done!
I also added the \`  \` around other steps.

mole99:
@nathanchandanson Thank you!

mole99 (approved):
LGTM

mole99:
@donn I added the `feat!` because this is a breaking change. Existing paths need to be converted to lists.
Was there anything else to do on a breaking change?

donn:
We really shouldn't break the API unless a feature was already discussed in an issue and maintainers agree there is no way to do it without breaking the API.

There are two possible avenues to do this without breaking the API:

1. Change type to:
```
Path | List[Path] | None
```
and update codepaths accordingly

2. Change it to `LINTER_VLTS` with a deprecated variable name `LINTER_VLT` by adding the following argument
```
deprecated_names=[("LINTER_VLT", lambda x: [x])],
```

Which do you prefer @mole99

mole99:
@donn I agree, I'd forgotten about how `deprecated_names` works ^^
Option 2 seems like the best to me. It keeps the documentation simple.

nathanchandanson:
@donn @mole99 I can do this tomorrow if you want.

Do you have examples/docs that show/explain how to add the `deprecated_names` ?

Edit: Nevermind I looked a bit into other steps and I found examples.
To do so I should add a variable as `deprecated_name` with the old name like you showed and then create a new one `LINTER_VLTS` with the définition I implemented ?

donn:
Update the existing variable to LINTER_VLTS and add the old name as a deprecated name

nathanchandanson:
Great thank you I'll do it tomorrow.

donn:
Thank you for your contribution, @nathanchandanson!

PR #984, opened 2026-07-17
feat: add Gdsfill.Filler step

dnltz:
Adds a step that wraps the gdsfill CLI to generate dummy metal fill, intended as an alternative to KLayout.Filler.


dnltz:
Depends on: https://github.com/fossi-foundation/nix-eda/pull/79 and the nix-eda tag `7.2.0` here in librelane.

PR #985, opened 2026-07-17
OpenROAD.GlobalPlacement: add option to enable GIF generation 

Essenceia:
Adding switch to enable global placement debug mode in librelane for generating global placement gif's. 

https://openroad.readthedocs.io/en/latest/main/src/gpl/README.html#debug-mode

Adding the following 2 configs : 
- `PL_GENERATE_GIF`: enable feature, disabled by default
- `PL_GENERATE_GIF_PAUSE`: number of steps between global routing iteration pauses, default set to a number large enough to never pause. 

Since this feature requires running openroad in gui mode, I am additionally adding a switch to force starting openroad in gui mode to the `OpenROADStep` parent's class `get_command`. 

This change has 2 main downsides: 
1. Since it requires launching the GUI people running of ssh without X forwarding will have there flow crash as openraod gui fails to launch. 
2. Running the global routing requires clicking the `Continue` button at the bottom left of the openroad gui, and closing the gui once finished for the flow to proceed. 

Example renders of expected results: 
- https://github.com/Essenceia/ethernet_switch_asic/blob/main/docs/gpl_placement.gif

Having fun with the renders: 
systolic_movement_placement


Essenceia:
P.S: ci is failing due to py formatting, looking into how I do that again.

donn:
in the `nix develop .#dev` you can just run `black .`

Essenceia:
Actually, seems like my modification of the parent class goes against the python coding standards, but fells to ugly to re-define get_command just for global routing as it risks missing certain additional parameters that the parent would have. 

I am honestly torn in what the best way of doing this would be. 

~For now I will re-define all the children `get_command` signature. Tell me if you wish for me to remove this change in the comments.~

I have stared into the abyss: it's too ugly, I can't do that, re-defining the `get_command` for global placement.

donn:
I did this for Difetto but I think I prefer overriding get_command entirely more lol

https://github.com/donn/difetto/blob/b1add985a1d40b6269107e4854bb386fb46a9435/librelane_plugin_difetto/steps.py#L442

librelane/steps/openroad.py
```diff
@@ -1586,9 +1586,34 @@ class _GlobalPlacement(OpenROADStep):
...
+            ),
         ]
     )
 
+    def get_command(self) -> List[str]:
+        enable_gui = self.config["PL_GENERATE_GIF"]
+        metrics_path = os.path.join(self.step_dir, "or_metrics_out.json")
+        return [
+            self.get_openroad_path(),
+            ("-gui" if enable_gui else "-exit"),
```
donn: I think for the best experience you can prepare a tuple as follows:

```python3
if os.getenv("_OPENROAD_GUI", "0") == "1":
  # Internal debug GUI flag
  mode_options = ("-gui",)
elif self.config["PL_GENERATE_GIF"]:
  # support GIF creation but still exit noninteractively
  mode_options = ("-gui", "-exit")
else:
  # default behavior
  mode_options = ("-exit",)
```

Then just splat it in place of the ternary condition:

```suggestion
            *mode_options,
```

This would also make `"\[INFO\] Please close the OpenROAD GUI so that the flow can proceed."` redundant.

Does this work?

Essenceia: Nice one, yes using the `-exit` argument works like a charm. 

Unfortunately since openroad really seems to hate duplicate arguments I reworked the tuple code a bit. eg: `-exit -exit`, as well as added white spaces.

Since there was just two additional arguments I handled duplicates manually, and didn't think it was worth the complexity of using a dictionary to do de-duplication.

donn:
- will take a look tomorrow or the day after, sorry for the wait

Essenceia:
No worry @donn, take all the time you need. Even if you only find time all week that is totally fine. :)

donn (approved):

PR #987, opened 2026-07-21
feat: make SYNTH_MUL_BOOTH default True

mlyoung101:
This makes SYNTH_MUL_BOOTH the default. 

Hopefully the bot should reply here with some metrics updates.

PR #988, opened 2026-07-22
feat: make SYNTH_MUL_BOOTH default True

mole99:
Copy of #987 to trigger the metrics bot


donn:
this bot sucks so much

PR #990, opened 2026-07-22
OpenROAD.RCX: Use set_extraction_rules_file

dnltz:
Use "set_extraction_rules_file" because "-ext_model_file" is now deprecated.

  [OpenROAD.RCX] [RCX-0514] The -ext_model_file option is deprecated.
  Use set_extraction_rules_file command instead.

### DIFF
--- a/librelane/scripts/openroad/rcx.tcl
+++ b/librelane/scripts/openroad/rcx.tcl
@@ -26,7 +26,6 @@ if { !$::env(RCX_MERGE_VIA_WIRE_RES) } {
 # RCX
 puts "Using RCX ruleset '$::env(RCX_RULESET)'…"
 define_process_corner -ext_model_index 0 CURRENT_CORNER
-log_cmd extract_parasitics $rcx_flags\
-    -ext_model_file $::env(RCX_RULESET)\
-    -lef_res
+set_extraction_rules_file $::env(RCX_RULESET)
+log_cmd extract_parasitics $rcx_flags -lef_res
 write_views


mole99:
@dnltz Thanks for the PR! We can do that once we've upgraded OpenROAD in our Nix flake. Currently, the CI fails because OpenROAD doesn't know `set_extraction_rules_file`.

dnltz:
> @dnltz Thanks for the PR! We can do that once we've upgraded OpenROAD in our Nix flake. Currently, the CI fails because OpenROAD doesn't know `set_extraction_rules_file`.

@mole99 ack - I was unsure whether this command is included in the tagged version. Do you have a label for "merge later when x"? 

It's also not critical since it's only a warning.

mole99:
No, but there's "to be merged" which is probably good enough.

I appreciate you fixing the warnings, as they usually distract from real errors :)

librelane/scripts/openroad/rcx.tcl
```diff
@@ -26,7 +26,6 @@ if { !$::env(RCX_MERGE_VIA_WIRE_RES) } {
 # RCX
 puts "Using RCX ruleset '$::env(RCX_RULESET)'…"
 define_process_corner -ext_model_index 0 CURRENT_CORNER
-log_cmd extract_parasitics $rcx_flags\
-    -ext_model_file $::env(RCX_RULESET)\
-    -lef_res
+set_extraction_rules_file $::env(RCX_RULESET)
```
donn, line 29: For situations like this if you'd prefer the merge to not wait for an OpenROAD upgrade which, historically speaking takes a lot of time and saps my mental willpower, you may choose to remain cross-compatible as follows:

```suggestion
if { [info procs set_extraction_rules_file] != "" } {
    set_extraction_rules_file $::env(RCX_RULESET)
    log_cmd extract_parasitics $rcx_flags -lef_res
} else {
    log_cmd extract_parasitics $rcx_flags\
        -ext_model_file $::env(RCX_RULESET)\
        -lef_res
}
```

dnltz: Totally fine for me to wait with this PR. Just want to make sure there is already a fix for this warning whenever OpenROAD will be updated.

PR #991, opened 2026-07-23
feat: allow to connect only some of the power/ground ports of an instance

mole99:
I noticed this issue when I wanted to connect only the VPWR/VGND ports of a power switch and leave the switched output unconnected for testing.

A simple fix: only throw an error if there are more than one connections, only make the connection if there is a connection.


mole99:
Gonna merge as discussed.

PR #994, opened 2026-07-24
fix: replace `KLAYOUT_*_OPTIONS` with `KLAYOUT_*_DEFINES`

mole99:
Fixes a number of corner cases resulting from ambiguity regarding the types of certain variables by ensuring all variables are defined in the type they are passed to KLayout to, i.e., string defines.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,22 @@ Style Notes
 
 -->
 
+# 3.0.9
+
+## Steps
+
+* `KLayout.DRC`
+
+  * Added `KLAYOUT_DRC_DEFINES`, which replaces `KLAYOUT_DRC_OPTIONS`.
+
+* `KLayout.Filler`
+
+  * Added `KLAYOUT_FILLER_DEFINES`, which replaces `KLAYOUT_FILLER_OPTIONS`.
+
+* `KLayout.Density`
+
+  * Added `KLAYOUT_DENSITY_DEFINES`, which replaces `KLAYOUT_DENSITY_OPTIONS`.
+
 # 3.0.8
 
 ## Steps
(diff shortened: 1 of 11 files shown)


mole99:
It seems there's another hidden issue/bug:

The sky130 PDK config currently has the following DRC settings:

```
set ::env(KLAYOUT_DRC_OPTIONS) [dict create beol 1 feol 1 floating_metal 0 seal 1 offgrid 1] ;
```

This incorrectly results in a single string for `KLAYOUT_DRC_OPTIONS`: `beol 1 feol 1 floating_metal 0 seal 1 offgrid 1`

By using `dict set`, this properly results in a dict on Python's side:

```
set ::env(KLAYOUT_DRC_OPTIONS) [dict create]
dict set ::env(KLAYOUT_DRC_OPTIONS) beol 1
dict set ::env(KLAYOUT_DRC_OPTIONS) feol 1
dict set ::env(KLAYOUT_DRC_OPTIONS) floating_metal 0
dict set ::env(KLAYOUT_DRC_OPTIONS) seal 1
dict set ::env(KLAYOUT_DRC_OPTIONS) offgrid 1
```


The reason for this is that we have overridden `dict set` and set the key/values in parallel in Python.
We cannot do the same for dict create since it does not name a variable.

I'm not really sure what the solution is here, except for updating the sky130 PDK config.

Ideally, instead of overriding Tcl functions, we would be able to access all the results and types(?) after the Tcl interpreter has finished.

donn:
@mole99 pdk_compat is designed for situations such as this

donn (approved):
After my changes things look like they work for me, but I'll leave it to you to merge

donn:
Something looks like it's still broken with KLAYOUT_DENSITY_OPTIONS but that's not really my area

Issue #995, opened 2026-07-26
ASAP7 Support


benreynwar:
### Description

It would be cool to have asap7 support in librelane to make it easy to do design space exploration with a more recent process.

### Proposal

I've been experimenting with getting asap7 working myself (with the help of an LLM, so factor that into how trustworthy these comments are).

I think what is needed is:
1) Get asap7 added to pdk_tools.  This involves copying over a bunch of files from orfs into pdk_tools, which feels a bit ugly, but I think is better than having pdk_tools grab them from the orfs repo as it builds the folder. (fossi-foundation/open-pdks#533)
2) Add librelane support for having the LEF/DEF layer mapping embedded in the .lyt file rather than as a separate .map file. (#999)
3) Add librelane support for having a PDK supplied tcl file for specifying the routing tracks.  ASAP7's track layout can't be represented in tracks.info. (#998)
4) Add librelane support for having a PDK supplied tcl file for specifying the PDN setup.  This means we could use the asap7-specific PDN tcl script from ORFS as the default for asap7 in librelane.  Maybe it's better to extract the parameters that librelane needs, and use the default librelane pdn, but I thought that sticking with the existing script from ORFS was probably more sensible. (#997)
5) Because the asap7 liberty files use quite different units than the PDKs that librelane has worked with so far this exposes some ambiguity about what units are used in different locations (#996)

PR #1000, opened 2026-07-28
pyosys: preserve Liberty file order

benreynwar:
Some PDKs split their standard cells across multiple Liberty files. ABC uses the first file as its base library, so the order is important.  This is triggered by the asap7 pdk. (#995)

Librelane should maintain the order of the libraries from the pdk.

### DIFF
--- a/librelane/common/toolbox.py
+++ b/librelane/common/toolbox.py
@@ -363,7 +363,7 @@ class Toolbox(object):
 
     def remove_cells_from_lib(
         self,
-        input_lib_files: FrozenSet[str],
+        input_lib_files: Tuple[str, ...],
         excluded_cells: FrozenSet[str],
     ) -> List[str]:
         """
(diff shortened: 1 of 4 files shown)


donn:
We don't usually take LLM changesets but this one in particular is fairly small and focused: can you add an Assisted-by: trailer to your PR body as per these instructions? https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#transparency

I'll have to discuss with the others.

benreynwar:
Makes sense.  I would totally understand if you don't want to merge LLM generated code.  Even making sure I'm following all the changes and code, it's definitely not the same as going through the process to understand the codebase properly myself and then making the changes myself.
It's probably worth updating the contributors guide.

donn:
Yes, I'm working on the document in question but haven't had the chance to finalize it yet.

benreynwar:

Issue #999, opened 2026-07-26
Make KLAYOUT_DEF_LAYER_MAP optional


benreynwar:
### Description

asap7 (#995) supplied this map embedded in the .lyt file, rather than as a separate .map file.  It would be nice for the map argument to be optional so that the embedded map could be used.

PR #1001, opened 2026-07-28
Support embedded KLayout LEF/DEF layer maps (LLM-aided)

benreynwar:
In the asap7 PDK the lef/def map in supplied embedded the .lyt file.
The resolves issue #999 and contributes to #995


benreynwar:

PR #1002, opened 2026-07-28
Support multiple track families per direction

### LINKED ISSUE #998: Support ASAP7 tracks
### Description

asap7 (#995) has a tracks layout that isn't consistent with the tracks.info format.
Instead the pdk could provide a tcl file that directly sets the tracks up.

### Proposal

_No response_

benreynwar:
Resolved #998
Contributes to #995 

Updates the parsing of the tracks.info file in the pdk so that we can have multiple families of tracks in each direction.
Necessary for asap7 support.


donn:
I'd prefer we either translate `make_tracks` in asap7 to an appropriate format or simply add an option to `source` a track file in OpenROAD. This is too complicated.

benreynwar:
I don't think this is as bad as it looks!

The tracks.info for asap7 that I'm using in the PDK is

...
M3 X 0.009 0.036
M3 Y 0.009 0.036
M2 X 0.009 0.036
M2 Y 0.045 0.270
M2 Y 0.081 0.270
M2 Y 0.117 0.270
M2 Y 0.153 0.270
M2 Y 0.189 0.270
M2 Y 0.225 0.270
M2 Y 0.270 0.270
M1 X 0.009 0.036
M1 Y 0.009 0.036

So really the only difference is that in the M2 layer has a bunch of tracks for Y.
All the change does it create a list for each (layer, cardinal) rather than a single entry.
Then it loops through the lists.

What makes it ugly is that the `make_tracks` command in OpenROAD does one X route and one Y route at the same time. This would be cleaner if OpenROAD supported separate make_tracks commands for the X direction and Y direction.

donn:
Reading it again… I'll let the CI run but I need to sit down and review it a bit more thoroughly at some point

benreynwar:

Issue #997, opened 2026-07-26
Add DEFAULT_PDN_CFG for a pdk.


benreynwar:
### Description

For asap7 (#995) it would be nice to be able to make the orfs pdn tcl script the default for constructing the pdn.  We would add the script to the open_pdks asap7 folder.

This would be an alternative to supplying all the parameters that librelane needs to build the pdn itself.

### Proposal

Or maybe it's better just to extract the parameters that librelane needs to build the pdn?  I'm not at all sure.  I thought adding the orfs pdn script was safer, since it keeps the librelane solution closer to the working orfs one.

donn:
No, I'm in agreement. In theory even, it could just be as simple as adding `pdk=True` to the existing `PDN_CFG`.

As you surmised, the tough part is variable support. In a theoretical packaging under ciel, we can include a modified version of ORFS's PDN script that supports the LibreLane variables.

benreynwar:
Is it important to have values for all the variables if there is a PDN tcl script that doesn't use them?

If it is important I can try to get values from them. I think the script is similar enough so that should be possible but I haven't looked carefully.

PR #1003, opened 2026-07-28
allow PDKs to provide PDN_CFG (LLM-aided)

benreynwar:
Make PDN_CFG a PDK variable and allow the generic PDN parameters to be omitted when a custom configuration is provided.

The bigger chance here is allowing the PDN parameters to be emitted when a PDN_CFG is provided.  We could instead extract the PDN parameters from the script, but I wasn't confident I could do that correctly.

#997, #995


donn:
This one is a bit too big of a changeset. I'd prefer dummy values for asap7 that are ignored.

benreynwar:

Issue #996, opened 2026-07-26
Unit Clarity


benreynwar:
### Description

It would be good to have more clear rules about what units for constants are used in which places.

As an example for GF180 the units of resistance in the liberty files are ohms, whereas librelane uses kiloohms as the standard units.  It's not clear which of these units constants use in different files.

For ASAP7 this becomes more important since the units of time in the liberty file are ps, while librelane uses ns as the standard, so timing constraints become ambiguous (#995)

### Proposal

I don't think it matters whether files or parameters are defined with respect to the liberty file units or the librelane standard units, but I do think that it needs to be clearer exactly which files use which units.

benreynwar:
I think this is going to be the hardest part of asap7 support, or at least I'm finding it the most difficult. I'm currently just trying to get my head around all the different places that constants are defined and what the implicit units are. I'll make a document once I think I know what is going on.

benreynwar:
In open-pdks/gf180mcu/librelane/config.tcl it looks like the LAYERS_RC seem to be defined in units of kOhm/um, but would be interpreted as Ohm/um by OpenRoad when the set_layer_rc is used.  Not currently an issue, since librelane pins ciel to an older open-pdks that doesn't have these values, but still an example of how it would be nice to be clearer about the units.

donn:
@mole99 potentially critical?

mole99:
@benreynwar Do you know if `set_layer_rc` always expects Ohm/um, or does that change when `set_units` is used? The documentation for `set_layer_rc` doesn't mention a unit, only "Resistance per unit length": https://openroad.readthedocs.io/en/latest/main/src/est/README.html#set-layer-rc

I obtained the RC constants for gf180mcu directly from OpenROAD-flow-scripts (they have been updated in the meantime): https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/platforms/gf180/setRC.tcl
It could be that ORFS uses `set_units` at some point earlier in the flow.

It's good that you mentioned this. In any case, the issue shouldn't be critical, as the RC constants are only used for estimation.

benreynwar:
The librelane documentation for the LAYERS_RC Variable says that the units are whatever is defined in the first lib file.

[...]

PR #1004, opened 2026-07-28
WIP: Fix unit handling at LibreLane tool boundaries (LLM-aided)

benreynwar:
This is a first attempt at being explicit about what units and getting used and setting the units explicitly before interacting OpenROAD. I leaned heavily on an LLM to find the places where changes were needed, and I'm not at all confident that the changes aren't breaking things.

Really I don't understand the codebase well enough to be making these changes, but they are enabling me to use asap7 locally so I thought it was worth creating this as a pull request at least as a starting point.

The main changes are:
 `set_sta_cmd_units` is being used before any openroad commands that assume values have the opensta units.
 `set_first_liberty_units` is used as a hacky way of supporting the LAYER_RC and VIA_R commands which use the units defined in the liberty files.

#996 #995

### DIFF
--- a/librelane/common/toolbox.py
+++ b/librelane/common/toolbox.py
@@ -513,6 +513,8 @@ class Toolbox(object):
     ) -> Optional[Decimal]:
         """
         Extract the voltage from the default operating conditions of a liberty file.
+        Liberty files with an explicit voltage unit other than one volt are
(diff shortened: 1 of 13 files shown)


benreynwar:
Closing. Will hopefully submit one without any LLM-generated content.

PR #1007, opened 2026-08-04
Fix curl command for Nix installation on macOS

maehne:
The current curl command for Nix installation on macOS causes a shell error.


donn (approved):
Thanks for the catch. How long has that been there…?

PR #1014, opened 2026-08-09
hotfix: potential integer division of TIME_DERATING_CONSTRAINT in SDC files

### LINKED ISSUE #1013: Timing derate is silently zeroed: integer division in base.sdc means set_timing_derate is always 1.0
Up front: an LLM found this while auditing our flow's constraint handling. I verified the behaviour by hand against 3.0.5 and current master before filing, and the evidence below is from runs on our own designs.

## Description

`TIME_DERATING_CONSTRAINT` never reaches OpenSTA. In `librelane/scripts/base.sdc` the percentage is divided by an integer literal, and the value itself arrives as an integer string, so Tcl does integer division and truncates to zero:

```tcl
# librelane/scripts/base.sdc:71-73 (master)
puts "\[INFO] Setting timing derate to: $::env(TIME_DERATING_CONSTRAINT)%"
set_timing_derate -early [expr 1-[expr $::env(TIME_DERATING_CONSTRAINT) / 100]]
set_timing_derate -late [expr 1+[expr $::env(TIME_DERATING_CONSTRAINT) / 100]]
```

The default is assigned as a Python int in `librelane/config/pdk_compat.py:328`:

```python
if "TIME_DERATING_CONSTRAINT" not in config:
    new["TIME_DERATING_CONSTRAINT"] = 5
```

so `$::env(TIME_DERATING_CONSTRAINT)` is the string `5`. In the OpenROAD Tcl interpreter shipped with 3.0.5:

```
puts "early=[expr 1-[expr 5 / 100]] late=[expr 1+[expr 5 / 100]]"
early=1 late=1

puts "early=[expr 1-[expr 5.0 / 100]] late=[expr 1+[expr 5.0 / 100]]"
early=0.95 late=1.05
```

Every run using the default therefore calls `set_timing_derate 1.0`, i.e. no derating at all.

It is easy to miss because line 71 echoes the environment variable rather than the computed multiplier, so the log reads `[INFO] Setting timing derate to: 5%` on runs where the applied derate is 1.0. `write_sdc` does not serialise derates either, so `final/sdc/*.sdc` gives no hint of the discrepancy.

## Expected Behavior

`TIME_DERATING_CONSTRAINT: 5` should produce `set_timing_derate -early 0.95` and `-late 1.05`.

## Impact

`base.sdc` is read by the resizer steps as well as the STA steps, so this affects optimisation and not only reporting: `repair_design` / `repair_timing` stop as soon as the underated numbers meet the constraint.

Two sky130A designs of ours, same routed databases, post-PnR STA re-run with the derate applied:

| design | as reported | derate applied |
| --- | --- | --- |
| A (~32 mm², 26 ns period), nom_ss setup WNS | -9.354 | -11.358 |
| B (~1.4 mm², 20 ns period), max_ss setup WNS | +0.800 | -0.351 |

Design B goes from meeting setup at all nine corners to missing one.

Rebuilding B with `TIME_DERATING_CONSTRAINT: 5.0` and nothing else changed returns it to +0.143 at all nine corners, 0 setup violations, LVS and DRC unchanged, at a cost of +181 instances (+0.27%). So fixing this is not purely a correctness-of-reporting change; it also lets the resizer do work it was previously skipping.

## Same pattern elsewhere

- `base.sdc:19-20` — `[expr $::env(CLOCK_PERIOD) * $::env(IO_DELAY_CONSTRAINT) / 100]`. With `CLOCK_PERIOD` 26 and the `pdk_compat` default `IO_DELAY_CONSTRAINT` 20 this yields 5 rather than 5.2.
- `librelane/scripts/openroad/common/io.tcl:98-99` — `IO_PCT` and `SYNTH_TIMING_DERATE` both become 0.
- `librelane/scripts/openroad/cts.tcl:86` — `CTS_DELAY_BUFFER_DERATE_PCT`.

Already correct, for reference: `base.sdc:61` (`/ 1000.0`), `cts.tcl:82` (`/ 100.0`), `gpl.tcl:42` (`/ 100.0`).

## Suggested fix

Float divisor at each affected site:

```tcl
set_timing_derate -early [expr 1-[expr $::env(TIME_DERATING_CONSTRAINT) / 100.0]]
set_timing_derate -late [expr 1+[expr $::env(TIME_DERATING_CONSTRAINT) / 100.0]]
```

Coercing these values to float when the environment is exported would be more robust, since it also covers any future `expr` site that divides a percentage.

One caveat worth a changelog note: designs that currently pass with small positive slack may start failing once the derate actually applies. That is the correct behaviour, but it will look like a regression to anyone upgrading.

Happy to open a PR if that is useful.

## Reproduction material

No design is required — the Tcl snippet above reproduces it directly in the bundled OpenROAD.

The flow-level symptom is visible in any existing run: `*-openroad-stapostpnr/*/sta.log` prints the configured percentage while the applied derate is 1.0. Setting `TIME_DERATING_CONSTRAINT: 5.0` in the config changes the log line to `5.0%` and changes the reported slacks; setting `5` does not.

## Environment report

```sh
kernel: Linux
kernel_version: 6.18.37
supported: True
distro: nixos
distro_version: 25.11
python_version: 3.13.9
librelane_version: 3.0.5 (nix flake github:librelane/librelane/3.0.5)
tkinter: False
container_info: no compatible container engine found in PATH (tried docker, podman)
nix_info:
  version_string: nix (Nix) 2.31.5
  nix_command: True
  flakes: True
pdk: sky130A (ciel 8afc8346a57fe1ab7934ba5a6056ea8b43078e71)
```

Source lines cited above were read from `librelane/librelane` master via the GitHub API on 2026-08-08, so they are current and not specific to the 3.0.5 tag we run.

donn:
TIME_DERATING_CONSTRAINT can be set as an integer and is then divided by 100 in Tcl's expr, which the result is always 0. A fix was added to the default SDC files, and mitigations were peppered throughout the rest of the codebase.

---

Resolves #1013

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,36 @@ Style Notes
 
 -->
 
+# 3.0.7
+
+## Steps
+
+* `OpenROAD.*`
+
+  * Fixed issue reported by [@Nexlab-One](https://github.com/Nexlab-One) where
+    `TIME_DERATING_CONSTRAINT` was integer-divided instead of float-divided, and
+    additionally `IO_DELAY_CONSTRAINT` if specified using old syntax (`IO_PCT`).
+
+## Misc. Enhancements/Bugfixes
+
+* `librelane.steps.TclStep.value_to_tcl`
+  * Fixed long-incorrect docstring.
+  * `Decimal` values that are integers are now always output with a `.0` to
+    pre-empt any further integer division issues (as well as backport the fix to
+    any user-written SDC files).
+
+* `librelane.config.pdk_compat`
+  * For extra resilience, `TIME_DERATING_CONSTRAINT`, `IO_DELAY_CONSTRAINTS` are
+    now both set as floats.
+
+# 3.0.6
+
+## Misc. Enhancements/Bugfixes
+
+* Slight internal rework of `librelane.common.metrics` that should not affect
+  its API, also fixing "folders" to "directories" in documentation where
+  appropriate.
+
 # 3.0.5
 
 ## Tool Updates
--- a/librelane/config/pdk_compat.py
+++ b/librelane/config/pdk_compat.py
@@ -325,9 +325,9 @@ def migrate_old_config(config: Mapping[str, Any]) -> Dict[str, Any]:
         if "CLOCK_TRANSITION_CONSTRAINT" not in config:
             new["CLOCK_TRANSITION_CONSTRAINT"] = 0.15
         if "TIME_DERATING_CONSTRAINT" not in config:
-            new["TIME_DERATING_CONSTRAINT"] = 5
+            new["TIME_DERATING_CONSTRAINT"] = 5.0
         if "IO_DELAY_CONSTRAINT" not in config:
-            new["IO_DELAY_CONSTRAINT"] = 20
+            new["IO_DELAY_CONSTRAINT"] = 20.0
 
         # Unspecififed is fine
         # if "FP_IO_MIN_DISTANCE" not in config:
--- a/librelane/examples/spm/src/impl.sdc
+++ b/librelane/examples/spm/src/impl.sdc
@@ -69,5 +69,5 @@ puts "\[INFO] Setting clock transition to: $::env(CLOCK_TRANSITION_CONSTRAINT)"
 set_clock_transition $::env(CLOCK_TRANSITION_CONSTRAINT) $clocks
 
 puts "\[INFO] Setting timing derate to: $::env(TIME_DERATING_CONSTRAINT)%"
-set_timing_derate -early [expr 1-[expr $::env(TIME_DERATING_CONSTRAINT) / 100]]
-set_timing_derate -late [expr 1+[expr $::env(TIME_DERATING_CONSTRAINT) / 100]]
+set_timing_derate -early [expr 1-[expr $::env(TIME_DERATING_CONSTRAINT) / 100.0]]
+set_timing_derate -late [expr 1+[expr $::env(TIME_DERATING_CONSTRAINT) / 100.0]]
--- a/librelane/examples/spm/src/signoff.sdc
+++ b/librelane/examples/spm/src/signoff.sdc
@@ -64,5 +64,5 @@ puts "\[INFO] Setting clock uncertainty to: 0.1"
 set_clock_uncertainty 0.1 $clocks
 
 puts "\[INFO] Setting timing derate to: $::env(TIME_DERATING_CONSTRAINT)%"
-set_timing_derate -early [expr 1-[expr $::env(TIME_DERATING_CONSTRAINT) / 100]]
-set_timing_derate -late [expr 1+[expr $::env(TIME_DERATING_CONSTRAINT) / 100]]
+set_timing_derate -early [expr 1-[expr $::env(TIME_DERATING_CONSTRAINT) / 100.0]]
+set_timing_derate -late [expr 1+[expr $::env(TIME_DERATING_CONSTRAINT) / 100.0]]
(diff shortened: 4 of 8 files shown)


donn:
@Nexlab-One could you verify that this PR works for your purposes?

mole99 (approved):
LGTM

Perhaps we could set the default values for the `TIME_DERATING_CONSTRAINT` and `IO_DELAY_CONSTRAINT` variables?
Then, we could probably remove the entries from pdk_compat.

donn:
!metrics

donn:
@mole99 We need to update the PDK at some point, yes

Nexlab-One:
> @Nexlab-One could you verify that this PR works for your purposes?

Yea, I can confirm it is working for my purpose, thanks for the quick fix, and for going further than the
report did.

The agent checked 3.0.7 as built (`github:librelane/librelane/3.0.7`, rev `77ec836`) rather than just the
diff, and evaluated the shipped expressions in OpenROAD's own Tcl.

The `value_to_tcl` change is the one the llm didn't think of and is probably the most valuable of the
fixes. Which was user-written SDC files that still divide by `100`, that we'd have kept hitting
silently.

## On the missing regression test
That LLM noticed regression tests missing while writing a test  and thought worth flagging: 
**the two fixes mask each other, so neither is currently guarded.**

Feeding `value_to_tcl(Decimal(5))` (i.e. `"5.0"`) through the *old* `/ 100` expression still yields
0.95 — the `tclstep.py` fix covers the SDC bug. So a test written the obvious way passes even if
`base.sdc` is reverted:

```
pre-3.0.7 expression, value_to_tcl(Decimal(5)) -> early=0.95  PASS   <- reverting base.sdc goes unnoticed
pre-3.0.7 expression, raw int "5"              -> early=1.0   FAIL   <- this one catches it
```

To guard each layer you need the SDC test to feed a **raw integer**, and a separate assertion on
`value_to_tcl`. Below is what I ended up with — it reads the shipped `base.sdc` rather than
hardcoding the expression, so reverting the file fails the test. Take it, rewrite it, or ignore it;
I've verified it passes on 3.0.7 and fails against the pre-3.0.7 expression.

```python
@pytest.mark.all
def test_sdc_time_derate_not_integer_divided():
    """
    Regression test for #1013.

    TIME_DERATING_CONSTRAINT was integer-divided in the SDC files, so an integer
    value silently produced a derate of 1.0 (no derating) while the log still
    printed the requested percentage.

    Feeds a RAW INTEGER deliberately: TclStep.value_to_tcl now emits "5.0" for an
    integral Decimal, which independently masks the bug, so passing the converted
    value here would let a revert of base.sdc go unnoticed.
    """
    import os
    import re
    import tkinter

    from librelane.common import get_script_dir

    sdc = os.path.join(get_script_dir(), "base.sdc")
    with open(sdc, encoding="utf8") as f:
        lines = [ln for ln in f.read().splitlines() if "set_timing_derate" in ln]
    assert lines, "no set_timing_derate lines found in base.sdc"

    tcl = tkinter.Tcl()
    tcl.eval("set ::env(TIME_DERATING_CONSTRAINT) 5")

    derates = {}
    for line in lines:
        match = re.match(r"\s*set_timing_derate\s+-(early|late)\s+\[(.+)\]\s*$", line)
        assert match is not None, f"could not parse: {line}"
        derates[match[1]] = float(tcl.eval(match[2]))

    assert derates["early"] == pytest.approx(0.95), "early derate not applied"
    assert derates["late"] == pytest.approx(1.05), "late derate not applied"


@pytest.mark.all
def test_value_to_tcl_integral_decimal_has_point():
    """
    Companion to the above: integral Decimals must not reach Tcl as bare
    integers, or any user SDC still dividing by 100 silently loses its derate.
    """
    from decimal import Decimal

    assert TclStep.value_to_tcl(Decimal(5)) == "5.0"
    assert TclStep.value_to_tcl(Decimal("2.5")) == "2.5"
```

The same shape would extend to `io.tcl`'s `IO_PCT` / `SYNTH_TIMING_DERATE` if you think it's worth
it — those fail identically on a raw integer.

<sub>Disclosure: the original bug was found by an LLM-assisted analysis of a timing discrepancy in
our own designs; the verification above and this test were written and checked an llm.</sub>

PR #1015, opened 2026-08-09
fix: `abc -fast` being used with Yosys ≥ 0.68

### LINKED ISSUE #1012: Synthesis fails with Yosys ≥ 0.68: librelane_synth calls the removed abc -fast
### Description

LibreLane smoketest fails with the latest Yosys version.

### Expected Behavior

No fail.

### Environment report

```sh
LibreLane: main (3f90ac49), also 3.0.6
Yosys: fails with 0.68; works with 0.67
Reproducer: any librelane flow with the pyosys Yosys synthesis step against Yosys ≥ 0.68 — it fails at the abc pass in Yosys.Synthesis before writing any netlist.
```

### Reproduction material

Just run the smoketest.

### Relevant log output

```sh
Not needed.
```

### LLM Policy Confirmation

- [x] Regardless of how this issue was found, bug report text was written and verified by a human.

tmarhguy:
- add new function called yosys_version_at_least which checks if the current yosys version is >= the version components, the latter passed as integer variable arguments
- only use `abc -fast` for Yosys versions < 0.68
- only use `plugin -i slang` for Yosys versions < 0.67

---

Fixes #1012

### DIFF
--- a/librelane/scripts/pyosys/synthesize.py
+++ b/librelane/scripts/pyosys/synthesize.py
@@ -38,7 +38,7 @@ from typing import List, Optional
 
 import click
 
-from ys_common import ys, yosys_version
+from ys_common import ys, yosys_version_at_least
 from construct_abc_script import ABCScriptCreator
 
 
@@ -226,7 +226,7 @@ def librelane_synth(
     librelane_opt(d, fast=True)
 
     abc_pass = ["abc"]
-    if yosys_version() < (0, 68):
+    if not yosys_version_at_least(0, 68):
         abc_pass.append("-fast")
     if abc_dff:
         abc_pass.append("-dff")
--- a/librelane/scripts/pyosys/synthesize.py
+++ b/librelane/scripts/pyosys/synthesize.py
@@ -225,9 +225,7 @@ def librelane_synth(
     librelane_opt(d, fast=True)
     librelane_opt(d, fast=True)
 
-    d.run_pass(
-        "abc", "-fast", *(["-dff"] if abc_dff else [])
-    )  # Run ABC with fast settings
+    d.run_pass("abc", *(["-dff"] if abc_dff else []))
     d.run_pass("opt", "-fast")  # MORE fast optimization
 
     # Checks and Stats
--- a/librelane/scripts/pyosys/synthesize.py
+++ b/librelane/scripts/pyosys/synthesize.py
@@ -38,7 +38,7 @@ from typing import List, Optional
 
 import click
 
-from ys_common import ys
+from ys_common import ys, yosys_version
 from construct_abc_script import ABCScriptCreator
 
 
@@ -225,7 +225,12 @@ def librelane_synth(
     librelane_opt(d, fast=True)
     librelane_opt(d, fast=True)
 
-    d.run_pass("abc", *(["-dff"] if abc_dff else []))
+    abc_pass = ["abc"]
+    if yosys_version() < (0, 68):
+        abc_pass.append("-fast")
+    if abc_dff:
+        abc_pass.append("-dff")
+    d.run_pass(*abc_pass)
     d.run_pass("opt", "-fast")  # MORE fast optimization
 
     # Checks and Stats
--- a/librelane/scripts/pyosys/ys_common.py
+++ b/librelane/scripts/pyosys/ys_common.py
@@ -45,7 +45,7 @@ _yosys_version_extracted: Optional[Tuple[int, int]] = None
 def yosys_version_at_least(*target: Tuple[int, ...]) -> Tuple[int, int]:
     """
     check if version is >= the target components, i.e.
-    
+
     yosys_version_at_least(0, 68) checks if the verison >= 0.68
     """
     global _yosys_version_extracted
--- a/librelane/scripts/pyosys/ys_common.py
+++ b/librelane/scripts/pyosys/ys_common.py
@@ -43,8 +43,8 @@ def yosys_version() -> Tuple[int, int]:
     raw = ys.yosys_maybe_version()
     if match := _YOSYS_VERSION_RX.search(raw):
         return (int(match.group(1)), int(match.group(2)))
-    ys.log(
-        "* Warning: Could not identify Yosys version. "
+    ys.log_warning(
+        "Could not identify Yosys version. "
         "Defaulting to the latest on all gated behavior.\n"
     )
     return (999, 999)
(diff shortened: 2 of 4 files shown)


tmarhguy:
CI is awaiting workflow approval for a first-time contributor. This is a one-line fix for #1012 (drops deprecated `abc -fast` for Yosys ≥ 0.68). Thanks!

mole99:
CI is now running. Thanks!

tmarhguy:
Smoketest and the vast majority of design tests (94/96) passed. The two failures (MS_DMAC_AHBL, zipdiv/gf180) look unrelated to the abc -fast change — happy to investigate if you think otherwise.

donn:
!metrics for my own curiosity

As a rule we try our best to support the latest tool version but we do not change any existing tool's behavior where possible. I'll look into gating this behavior.

> The two failures (MS_DMAC_AHBL, zipdiv/gf180) look unrelated to the abc -fast change

Only directly: any change to synthesis (especially abc) is re-rolling the dice on the netlist, which affects the rest of the flow.

librelane/scripts/pyosys/synthesize.py
```diff
@@ -225,9 +225,7 @@ def librelane_synth(
     librelane_opt(d, fast=True)
     librelane_opt(d, fast=True)
 
-    d.run_pass(
-        "abc", "-fast", *(["-dff"] if abc_dff else [])
-    )  # Run ABC with fast settings
+    d.run_pass("abc", *(["-dff"] if abc_dff else []))
```
donn: Just to be clear, this would change the behavior significantly in `main` and needs to be gated. I can get to that sometime this week, it'd use something like this:

```python3
import re
from typing import Tuple
from pyosys import libyosys as ys

def yosys_version() -> Tuple[int, ...]:
    version_search_rx = re.compile(r"^\s*Yosys (\d+)\.(\d+)(\+\d+)?")
    raw = ys.yosys_maybe_version()
    if match := version_search_rx.search(raw):
        return tuple(int(v) for v in match.groups() if v is not None)
    ys.warn("* Could not identify Yosys version. Defaulting to the latest on all gated behavior.\n")
    return (999, 999)

if yosys_version() >= (68, 0):
    ...
else:
    ...
```

But I'm too tired to test it atm

donn (changes requested):

librelane/scripts/pyosys/synthesize.py
tmarhguy: Thanks @donn  — pushed a gated version based on your sketch. `abc -fast` is only omitted on Yosys >= 0.68; older toolchains keep the existing behavior. Used `(0, 68)` as the threshold since Yosys versions are 0.x.

donn:
Would you be able to test if this works with later versions of Yosys? That's the part I'm too tired to do atm haha

tmarhguy:
Thanks @donn — I'll test the ≥ 0.68 path locally via nix develop + smoketest and report back here. certainly get some rest : ), will work on this!

librelane/scripts/pyosys/ys_common.py
```diff
@@ -36,6 +39,17 @@
         exit(-1)
 
 
+def yosys_version() -> Tuple[int, int]:
+    raw = ys.yosys_maybe_version()
+    if match := _YOSYS_VERSION_RX.search(raw):
+        return (int(match.group(1)), int(match.group(2)))
+    ys.log(
```
donn:
```suggestion
    ys.log_warning(
```

librelane/scripts/pyosys/ys_common.py
```diff
@@ -36,6 +39,17 @@
...
+        "* Warning: Could not identify Yosys version. "
```
donn:
```suggestion
        "Could not identify Yosys version. "
```

donn (commented):
also why not just use log_warning?

tmarhguy:
@donn  — local test update:

Ran `nix develop .#dev` + `python3 -m librelane --smoke-test` on this branch — smoketest passed.

Devshell Yosys is 0.62, so this exercises the gated legacy path (`abc -fast` kept for < 0.68). I couldn't get 0.68+ locally (devshell pins 0.62), so I'm relying on CI smoketest for the ≥ 0.68 branch once workflows are re-approved.

Also addressed the `log_warning` suggestion in e9bfba4

mole99:
@donn Since we are planning to do the next LibreLane release soon, we could just update Yosys in [nix-eda](https://github.com/fossi-foundation/nix-eda) to 0.68 and bump nix-eda in current dev? Then we wouldn't need any special handling around `-fast`.

Though beware, Yosys recently changed their build system, so it won't be as easy as just updating the version number.

donn:
@mole99 I have the code for the new build system in a branch but I don't have the opportunity to test it today because my ISP is cooked (using mobile data for this)…

mole99:
Oh, nice! (Well, except for the part about your ISP...) Let me know if you want me to rebase the branch and give it a try.

librelane/scripts/pyosys/ys_common.py
donn: One more favor: starting 0.68, slang is built-in, so this command would fail:

```suggestion
        if yosys_version() < (0, 67):
          self.run_pass("plugin", "-i", "slang")
```

Would appreciate it if you can get this too

donn:
@mole99 fyi https://github.com/fossi-foundation/nix-eda/pull/85

librelane/scripts/pyosys/ys_common.py
donn: EDIT: slang is included starting in 0.67, not 8. mb

hpretl:
Any chance for a merge in the next days? I would like to freeze the next release of IIC-OSIC-TOOLS this week. If not no worries then it goes in the next time.

donn:
@mole99 I decided to take this one over, your review would be 'preciated

mole99 (approved):
Looks all right to me. Nice also that you fixed slang with newer Yosys versions.

donn (approved):

hpretl:
@donn @mole99 Thanks for the fast merge! Since we are already on the `dev` branch (`3.1.0.dev2` to be precise), could you please also merge it there?

Correction: I see everything is merged in the `dev` branch; just a release seems to be missing. It would be cool to release a `dev3` with all the latest changes.

mole99:
@hpretl Thanks for reporting the original issue!

I have bumped the version to `3.1.0.dev3`. Though, the CI will take a while until it's published.

hpretl:
@mole99 Thanks Leo!

mole99:
Need to fix one more thing in the lock file, but then it should be good to go.

PR #1023, opened 2026-09-02
fix: handle OpenSTA scenes bridged as objects in io.tcl

jalcim:
Since OpenSTA 998806cb (2026-06-30), `sta::scenes` returns SWIG objects instead of names, and `set_cmd_scene` only accepts the object, whereas `worst_slack -corner` only accepts the name: `io.tcl` assumed strings everywhere, causing the pre-PnR STA step to fail with `_p_Scene is not the name of a scene`.

A `scene_name` helper (guarded by `is_object`, then `get_property name`) feeds `get_corner_names` and `get_corner_dict`, the latter mapping each name to whatever `sta::find_scene` returns; `set_sta_cmd_corner` now prefers `sta::set_scene`, which accepts a name across the entire OpenSTA 3 series.

Compatible with both the February 2026 OpenSTA pinned by librelane and the July one.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -25,6 +25,15 @@ Style Notes
 
 -->
 
+# 3.0.12
+
+## Steps
+
+* `OpenROAD.*`
+
+  * Added support for OpenSTA 3.1 in a backwards-compatible manner (Thanks,
+    [@jalcim](https://github.com/jalcim)!)
+
 # 3.0.11
 
 ## Tool Updates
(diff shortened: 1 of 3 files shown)


donn:
Hiya, thanks for the contribution. Just going to test against OpenSTA 3.0 as well and will approve if it works. :)

donn (approved):

donn:
Tested, working with all versions of OpenSTA I care about. Thank you so much!

…io.tcl is just going to become an OpenROAD compat layer at this rate…

PR #1024, opened 2026-09-03
[low priority] repair_design: Add missing variable + argument update

Essenceia:
While comparing against the openroad doc I noticed a few minor issues with the post gpl and grt scripts, this PR is my small fix. 

Changes: 
- Exposed the `*_MAX_UTILIZATION` variables, this variable was referenced in the tcl scripts but never set anywhere. 
- Removed the `-buffer_gain` argument from `repair_design` as it has been deprecated: https://openroad.readthedocs.io/en/latest/main/src/rsz/README.html#repair-design
- Replace use of `-buffer_gain` with `-pre_placement` from post gpl, similarly to what is already done on the openroad side: (https://github.com/The-OpenROAD-Project/OpenROAD/blob/2ac546acaf350111732655ae402ea952d06d362c/src/rsz/src/Resizer.tcl#L164-L167). 
- Remove `-buffer_gain` with no substitution for the post grl repair. Didn't think calling `-pre-placement` post routing sounded like a good idea. 

Notes: 
- Since `*_BUFFER_GAIN` was also never exposed to the config there is no risk of breaking anyone's configs. 

Things I am less certain of:
- Is there a need to invoke `-pre-placement` post grl ?


Essenceia:
Ah ... forgot to run the python syntax tools again

Issue #1028, opened 2026-09-05
PAD_LIBS declared as Dict but iterated flat in io.tcl; PAD_LIBS: null then silently disables all timing analysis


gonsolo:
## Summary

`PAD_LIBS` is declared as `Dict[str, List[Path]]` but consumed in Tcl as a flat list, so its corner-pattern **keys** are passed to `read_liberty` as if they were file paths.

The practical consequence is worse than the error itself: the natural workaround (`PAD_LIBS: null`) leaves the clock input pad with no timing model, so the clock never propagates into the core, STA analyses **zero paths**, and the flow then reports **zero setup/hold violations across all corners**. That reads as a clean signoff and is in fact a vacuous one.

Verified present in 3.0.8 (installed) and on `main` at time of writing (3.0.11 is the latest release).

## The mismatch

`librelane/config/flow.py:459-464`:

```python
Variable(
    "PAD_LIBS",
    Optional[Dict[str, List[Path]]],
    "A map from corner patterns to a list of associated liberty files. ...",
    pdk=True,
),
```

`librelane/scripts/openroad/common/io.tcl:239` (in `read_timing_info`):

```tcl
if { [info exists ::env(PAD_LIBS) ] } {
    foreach lib $::env(PAD_LIBS) {
        puts "Reading gpio pad timing for the '$corner_name' corner at '$lib'…"
        read_liberty -corner $corner_name $lib
    }
}
```

and again at `io.tcl:347`:

```tcl
if { [info exists ::env(PAD_LIBS) ] } {
    foreach pad_lib $::env(PAD_LIBS) {
        puts "Reading gpio pad timing library for the '$corner_name' corner at '$pad_lib'…"
        read_liberty -corner $corner_name $pad_lib
    }
}
```

A Tcl dict serialises as an interleaved key/value list, so `foreach` walks the corner globs as well as the paths:

```
Error: cannot read file *_tt_025C_5v00.
```

The asymmetry is clear against its neighbours in the same proc:

| Variable | Declared type | How Tcl consumes it |
|---|---|---|
| `LIB` | `Dict[str, List[Path]]` | resolved to `_CURRENT_CORNER_LIBS` via `toolbox.filter_views` first ✅ |
| `EXTRA_LIBS` | `List[Path]` | flat `foreach` — type matches ✅ |
| **`PAD_LIBS`** | **`Dict[str, List[Path]]`** | **flat `foreach` — type does not match ❌** |

[...]

PR #1029, opened 2026-09-05
fix: resolve PAD_LIBS per corner instead of iterating the dict flat

gonsolo:
Fixes the `PAD_LIBS` handling described in #1028.

## Problem

`PAD_LIBS` is declared `Optional[Dict[str, List[Path]]]` (corner pattern → libs) in `config/flow.py`, but `scripts/openroad/common/io.tcl` iterates it with a flat `foreach`. A Tcl dict serialises as an interleaved key/value list, so the corner globs themselves get passed to `read_liberty`:

```
Error: cannot read file *_tt_025C_5v00.
```

The neighbouring variables in the same proc show the asymmetry:

| Variable | Declared type | Tcl consumption |
|---|---|---|
| `LIB` | `Dict[str, List[Path]]` | resolved to `_CURRENT_CORNER_LIBS` via `filter_views` first ✅ |
| `EXTRA_LIBS` | `List[Path]` | flat `foreach` — type matches ✅ |
| `PAD_LIBS` | `Dict[str, List[Path]]` | flat `foreach` — type does not match ❌ |

## Why it's worth fixing rather than documenting

The natural workaround is `PAD_LIBS: null`, and it fails silently in the dangerous direction. With no liberty for the clock pad cell, `create_clock` on the pad port has nothing to propagate through, so STA analyses **zero paths** and then reports **zero setup/hold violations in every corner** — a summary indistinguishable at a glance from a design that genuinely closes timing.

## Change

Resolve `PAD_LIBS` per corner the same way `LIB` already is, using the existing `Toolbox.filter_views`, and expose it to Tcl pre-filtered as `_CURRENT_CORNER_PAD_LIBS`. Both `io.tcl` call sites updated. 13 lines.

## Verification

Tested on a gf180mcu chip-with-padring design (46 bidir + 4 input pads, ~12k registers), running the full flow:

| Configuration | Result |
|---|---|
| before, `PAD_LIBS` dict | `Error: cannot read file *_tt_025C_5v00.` |
| before, `PAD_LIBS: null` | `11987 unclocked register/latch pins`, `No paths found` |
| **after, `PAD_LIBS` dict** | `Reading gpio pad timing for the 'nom_tt_025C_3v30' corner at '…gf180mcu_fd_io__tt_025C_3v30.lib'`, **4004 paths clocked by `clk_PAD`, 0 unclocked warnings** |

The corner filtering is exercised, not bypassed: a three-entry `PAD_LIBS` dict correctly selects the `tt_025C_3v30` pad liberty for the `nom_tt_025C_3v30` corner.

Note the patch was developed against `main` and validated by running the same change against the installed 3.0.8 (identical code in both), since that matched the dependency set available locally.

### DIFF
--- a/librelane/scripts/openroad/common/io.tcl
+++ b/librelane/scripts/openroad/common/io.tcl
@@ -236,8 +236,8 @@ proc read_timing_info {args} {
         }
     }
     
-    if { [info exists ::env(PAD_LIBS) ] } {
-        foreach lib $::env(PAD_LIBS) {
+    if { [info exists ::env(_CURRENT_CORNER_PAD_LIBS) ] } {
+        foreach lib $::env(_CURRENT_CORNER_PAD_LIBS) {
             puts "Reading gpio pad timing for the '$corner_name' corner at '$lib'…"
             read_liberty -corner $corner_name $lib
         }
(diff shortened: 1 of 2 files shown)


donn:
Superseded by #1031.

PR #1033, opened 2026-09-08
feat: update Ciel and enable ihp-sg13cmos5l in test sets

mole99:
WIP


donn (approved):
I did my tweak, A-OK with this if you're okay with it

mole99:
Only fetching the libraries we actually need? I'm A-OK with this too ;)

PR #1036, opened 2026-09-10
Yosys.*Synthesis: support abc_new

donn:
- add experimental support for abc_new using `SYNTH_ABC_NEW` variable (default: false)
  - disables all abc strategy options other than `SYNTH_ABC_STRATEGY_SCRIPT` if set
- lib files are no longer trimmed for Yosys, -dont_use is passed to both abc and abc_new
- fix small issue in #937 where OpenROAD threads would be set to "None"
- fix abc sdc file setting driving cell incorrectly
- redocument every SYNTH_ABC variable to include their interactions

### DIFF
--- a/librelane/steps/pyosys.py
+++ b/librelane/steps/pyosys.py
@@ -268,7 +268,7 @@ class PyosysStep(Step):
         env["PYTHONPATH"] = ":".join(
             (env.get("PYTHONPATH", ""), os.path.join(get_script_dir(), "pyosys"))
         )
-        subprocess_result = super().run_subprocess(cmd, env=env, **kwargs)
+        subprocess_result = self.run_subprocess(cmd, env=env, **kwargs)
         return {}, subprocess_result["generated_metrics"]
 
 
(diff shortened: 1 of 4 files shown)


donn:
!metrics

donn:
!metrics

librelane/steps/pyosys.py
```diff
@@ -338,19 +338,27 @@ def get_command(self, state_in: State) -> List[str]:
...
+
+        libs_synth = [str(p) for p in scl_lib_list + pad_lib_list]
         extra_path = os.path.join(self.step_dir, "extra.json")
         with open(extra_path, "w") as f:
-            json.dump({"blackbox_models": blackbox_models, "libs_synth": libs_synth}, f)
+            json.dump(
+                {
+                    "blackbox_models": blackbox_models,
+                    "libs_synth": libs_synth,
+                    "excluded_cell_patterns": list(excluded_cell_patterns),
```
donn, line 358: Before I forget, this should also be passed to dfflibmap.

mole99 (approved):
Looks generally good to me!

I have a few questions about `abc_new`: what's so new about it, and why doesn't it require an abc script ^^

donn:
@mole99 @ravenslofty who maintains the abc pass asked me to do this because long-term Yosys is switching to the `abc_new` pass and it's better to get to the issues early, of which we found some. It's basically a completely rewritten pass that uses abc9 by default.

The reason the scripts aren't supported is that I'm too stupid/lazy to rewrite the default scripts to support abc9 atm.

